PR #8, opened 2018-04-25
Swap to sandbox when appstore advises

markrichardsg:


commonlambda/src/main/scala/com/gu/mobilepurchases/shared/lambda/DelegatingLambda.scala
```diff
@@ -69,20 +69,20 @@ class DelegatingLambda(
...
+
     val eventualResponses: Future[Seq[LambdaResponse]] = Future.sequence(Seq(futureLambdaResponse, futureDelegateResponse))
     delegateResponseF(lambdaRequest).transform((triedDelegateResponse: Try[LambdaResponse]) => {
       promiseDelegateResponse.complete(triedDelegateResponse)
       triedDelegateResponse
     })
-    Future {
-      underTest.apply(lambdaRequest)
-    }.transform((triedLambdaResponse: Try[LambdaResponse]) => {
+    Future.successful { underTest.apply(lambdaRequest) }.transform((triedLambdaResponse: Try[LambdaResponse]) => {
```
NathanielBennett: I like transform and transformWith. A lot

markrichardsg: 👍 it's pretty nice that the Try representation neatly fits with promise here.
Noticed Future.succesful shouldn't have been added, so reverted that

PR #10, opened 2018-04-26
Only return transactions requested

markrichardsg:


NathanielBennett:
:+1: Couple cosmetic things but looks good

PR #24, opened 2018-05-17
Read quantity metric

markrichardsg:


iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/purchases/UserPurchases.scala
```diff
@@ -23,10 +26,16 @@ class UserPurchasesImpl(
       .filter((_: String).startsWith("vendorUdid~"))
       .map(userPurchasePersistence.read(_: String, userPurchasesRequest.appId))
       .flatMap {
-        case Success(userPurchasesByUserIdAndAppId) => userPurchasesByUserIdAndAppId.map((_: UserPurchasesByUserIdAndAppId).purchases).getOrElse(Set())
+        case Success(userPurchasesByUserIdAndAppId) => {
+          userPurchasesByUserIdAndAppId.map((userPurchasesByUserIdAndAppId: UserPurchasesByUserIdAndAppId) => {
```
alexduf: Usually when you have a double arrow followed by an opening bracket it's a pretty good sign you should extract it in a smaller self-contained method with a nice name

iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/purchases/UserPurchases.scala
```diff
@@ -23,10 +26,16 @@ class UserPurchasesImpl(
...
+            val purchases: Set[UserPurchase] = userPurchasesByUserIdAndAppId.purchases
+            cloudWatchMetrics.queueMetric("purchases-quantity", purchases.size, StandardUnit.Count)
```
alexduf: Just so that I understand, how's that helping us ensuring our responses are matching the legacy service? Shouldn't there be some date comparison somewhere?
Also unless I've missed something we will have a big metric count on a graph but I can't see it's value in our context.

iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/lambda/DelegateUserPurchasesLambda.scala
```diff
@@ -87,6 +91,47 @@ class DelegateUserPurchasesLambdaComparator(cloudWatch: CloudWatch) extends Dele
...
+      logger.warn("Delegate Older Than Lambda: Delegate:\n {} Lambda\n: {}", delegatePurchaseSet: Any, lambdaPurchaseSet: Any)
+    }
+
+    def latestExpiryDate(purchases: Set[UserPurchase]): Option[String] = {
+      purchases.toSeq.filter(_.activeInterval.end < now).sortBy(_.activeInterval.end).lastOption.map(_.activeInterval.end)
+    }
+
+    val maybeLatestDelegatePurchaseExpiry: Option[String] = latestExpiryDate(delegatePurchaseSet)
+    val maybeLatestLambdaPurchaseExpiry: Option[String] = latestExpiryDate(lambdaPurchaseSet)
+    maybeLatestDelegatePurchaseExpiry match {
```
alexduf: I think this would be much more readable
```scala
(maybeLatestDelegatePurchaseExpiry, maybeLatestLambdaPurchaseExpiry) match {
  (Some(latestDelegate), Some(latestLambda)) => ...
  (Some, None) => ...
}
```

alexduf:
Also you may as well revert the first commit, unless you really want to track that.

iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/purchases/UserPurchases.scala
markrichardsg: Refactored

markrichardsg:
> Also you may as well revert the first commit, unless you really want to track that.

When the Delegate functionality is removed, so will many of the counts, this seems worth keeping.

iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/lambda/DelegateUserPurchasesLambda.scala
```diff
@@ -85,6 +92,42 @@ class DelegateUserPurchasesLambdaComparator(cloudWatch: CloudWatch) extends Dele
       }
     }
 
+  }
+  def latestMatched: Unit = {
```
alexduf: That would be nice to rename it as being `logLatestMatched()` (with the parenthesis) as it is a function with a side effect and to match the other functions

iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/lambda/DelegateUserPurchasesLambda.scala
```diff
@@ -85,6 +92,42 @@ class DelegateUserPurchasesLambdaComparator(cloudWatch: CloudWatch) extends Dele
...
+    cloudWatch.queueMetric("latest-matched", 1, StandardUnit.Count)
+  }
+
+  private def compareLatestPurchases(request: LambdaRequest, lambdaPurchaseSet: Set[UserPurchase], delegatePurchaseSet: Set[UserPurchase]): Unit = {
+    val now: String = ZonedDateTime.now.format(UserPurchase.instantFormatter)
+    def latestLambdaNewerThanDelegate: Unit = {
```
alexduf: Same here, it's a method with a side effect => `()` and all your other cloudwatch methods start with `log`

iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/lambda/DelegateUserPurchasesLambda.scala
alexduf: ~function~ method

PR #29, opened 2018-05-22
Remove unused app store receipt fields and treat rest as optional

markrichardsg:


iosvalidatereceipts/src/main/scala/com/gu/mobilepurchases/validate/ValidateReceiptsTransformAppStoreResponse.scala
```diff
@@ -40,16 +40,24 @@ class ValidateReceiptsTransformAppStoreResponseImpl extends ValidateReceiptsTran
...
-        case AutoRenewableSubsStatusCodes.valid | AutoRenewableSubsStatusCodes.ReceiptValidButSubscriptionExpired => ValidatedTransaction(
-          receipt.transaction_id,
-          validated = true,
-          finishTransaction = true,
-          Some(ValidatedTransactionPurchase(
-            receipt.product_id,
-            receipt.web_order_line_item_id,
+        case AutoRenewableSubsStatusCodes.valid | AutoRenewableSubsStatusCodes.ReceiptValidButSubscriptionExpired => {
+          val maybeValidatedTransactionPurchase: Option[ValidatedTransactionPurchase] = for {
+            product_id <- receipt.product_id
```
alexduf: I can see how for parsing reasons you have `product_id` in the case class but that would be nice to keep the camelCase for variable names here.

```scala
productId <- receipt.product_id
```

PR #30, opened 2018-05-30
Prioritise lambda response and if there is a failure by either, error…

markrichardsg:
… in the hope that the app re-sends validation

### DIFF
--- a/iosvalidatereceipts/src/main/scala/com/gu/mobilepurchases/lambda/DelegatingValidateReceiptLambda.scala
+++ b/iosvalidatereceipts/src/main/scala/com/gu/mobilepurchases/lambda/DelegatingValidateReceiptLambda.scala
@@ -53,7 +53,7 @@ class DelegatingValidateReceiptCompators(cloudWatch: CloudWatch) extends Delegat
           logLambdaExtras(0)
           logDelegateExtras(0)
           logReturnedTransactions(lambdaValidateResponse.transactions.size)
-          delegateResponse
+          lambdaResponse
         } else {
           val lambdaTransactions: Set[ValidatedTransaction] = lambdaValidateResponse.transactions
           val delegateTransactions: Set[ValidatedTransaction] = delegateValidateResponse.transactions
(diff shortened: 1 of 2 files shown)


iosvalidatereceipts/src/main/scala/com/gu/mobilepurchases/lambda/DelegatingValidateReceiptLambda.scala
```diff
@@ -64,25 +64,26 @@ class DelegatingValidateReceiptCompators(cloudWatch: CloudWatch) extends Delegat
...
+
           }
         }
 
       }
       case (Some(validateResponse), _) => {
         logger.warn(s"Validate mismatch for Request: $lambdaRequest \nLambda Response: $lambdaResponse \nDelegate Response: $delegateResponse")
         logDelegateOnly(validateResponse)
-        lambdaResponse
+        delegateResponse
```
alexduf: You're returning None here, why?

iosvalidatereceipts/src/main/scala/com/gu/mobilepurchases/lambda/DelegatingValidateReceiptLambda.scala
```diff
@@ -64,25 +64,26 @@ class DelegatingValidateReceiptCompators(cloudWatch: CloudWatch) extends Delegat
...
       }
       case (_, Some(validateResponse)) => {
         logger.warn(s"Validate mismatch for Request: $lambdaRequest \nLambda Response: $lambdaResponse \nDelegate Response: $delegateResponse")
         logLambdaOnly(validateResponse)
-        delegateResponse
+        lambdaResponse
```
alexduf: Same here

iosvalidatereceipts/src/main/scala/com/gu/mobilepurchases/lambda/DelegatingValidateReceiptLambda.scala
markrichardsg: If the validation succeeded in lambda but not delegate or vice versa, then there will be inconsistency in the purchases stored (eg: app store timeout)
If the app receives a bad response then the transactions should be re-sent by the device in the future and hopefully fix this inconsitency.

alexduf: Got it

PR #33, opened 2018-06-11
Log purchase quantity when empty

markrichardsg:


iosuserpurchases/src/main/scala/com/gu/mobilepurchases/userpurchases/purchases/UserPurchases.scala
```diff
@@ -35,11 +35,8 @@ class UserPurchasesImpl(
   }
 
   private def countAndExtractPurchaseSet(userPurchasesByUserIdAndAppId: Option[UserPurchasesByUserIdAndAppId]): Set[UserPurchase] = {
-    userPurchasesByUserIdAndAppId.map((userPurchasesByUserIdAndAppId: UserPurchasesByUserIdAndAppId) => {
-      val purchases: Set[UserPurchase] = userPurchasesByUserIdAndAppId.purchases
-      cloudWatchMetrics.queueMetric("purchases-quantity", purchases.size, StandardUnit.Count)
-      purchases
-    }).getOrElse(Set())
-
+    val purchases: Set[UserPurchase] = userPurchasesByUserIdAndAppId.map((_: UserPurchasesByUserIdAndAppId).purchases).getOrElse(emptyPurchases)
```
alexduf: There's no need to put the type everywhere, as much as I like types for readability on method's signatures, I think here you can drop it like so:
```scala
val purchases: Set[UserPurchase] = userPurchasesByUserIdAndAppId.map(_.purchases).getOrElse(emptyPurchases)
```

alexduf:
I was referring to the other type, but don't worry just push it :)

PR #42, opened 2019-05-24
Add basic lambda for obtaining/refreshing Play Developer API access tokens

jacobwinch:
This PR adds a new lambda which will be used to refresh (and store) access tokens used to authenticate against the [Play Developer API](https://developers.google.com/android-publisher/). Obviously this is not the finished code, but I wanted to get a pull request open before it gets too large!

The lambda is triggered by a CloudWatch Event Rule (on a schedule). The Cloudformation and lambda are both deployable using RiffRaff.

Next steps (for a subsequent PR):

- [ ] Add S3 permissions for storing the access tokens
- [ ] Add code to store access tokens in S3

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -231,9 +231,10 @@ Resources:
       MemorySize: 512
       Timeout: 45
       Events:
-        Type: Schedule
-        Properties:
-          Schedule: rate(15 minutes)
+        Schedule:
+          Type: Schedule
+          Properties:
+            Schedule: rate(15 minutes)
       Tags:
         Stage: !Ref Stage
         Stack: !Ref Stack
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -79,6 +79,13 @@ Resources:
                 - ssm:GetParametersByPath
               Effect: Allow
               Resource: !Sub arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${App}-iosvalidatereceipts/${Stage}/${Stack}
+        - PolicyName: googleoauth-config
+          PolicyDocument:
+            Statement:
+              Action:
+                - ssm:GetParametersByPath
+              Effect: Allow
+              Resource: !Sub arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${App}/${Stage}/google-oauth-lambda
         - PolicyName: dynamo
           PolicyDocument:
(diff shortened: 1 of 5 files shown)


cloudformation.yaml
```diff
@@ -199,6 +206,33 @@ Resources:
...
+          Stack: !Ref Stack
+          App: !Ref App
+      Description: Fetches access tokens for the Google Play Developer API
+      MemorySize: 512
+      Timeout: 45
+      Events:
+        Schedule:
+          Type: Schedule
+          Properties:
+            Schedule: rate(15 minutes)
```
jacobwinch: This schedule is up for debate. Tokens last for an hour, so running every 15 mins means that we should get a couple of retry attempts if there are intermittent problems with Google or AWS.

googleoauth/src/main/scala/com.gu.mobilepurchases.googleoauth/lambda/GoogleOAuth.scala
```diff
@@ -0,0 +1,43 @@
...
+
+object GoogleOAuth {
+
+  def accessToken(): Unit = {
+
+    val logger = LogManager.getLogger
+
+    val credentials = GoogleCredentials
+      .fromStream(new ByteArrayInputStream(fetchConfiguration.getString("google.serviceAccountJson").getBytes))
+      .createScoped("https://www.googleapis.com/auth/androidpublisher")
```
jacobwinch: This is the scope we need for the [API call](https://developers.google.com/android-publisher/api-ref/purchases/subscriptions/get#auth) I'm interested in for now. We may need to add further scopes here depending on what you need @alexduf

googleoauth/src/main/scala/com.gu.mobilepurchases.googleoauth/lambda/GoogleOAuth.scala
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+
+  def handler(input: InputStream, output: OutputStream, context: Context): Unit = {
+    accessToken()
+  }
+
+}
+
+object Debug extends App {
```
jacobwinch: This just allows us to run easily locally whilst this is under development

cloudformation.yaml
alexduf: Sounds sensible

PR #43, opened 2019-05-30
Add lambda to check subscription status via Play Developer API

jacobwinch:
This PR:

- Adds logic to check subscription status via the Play Developer API
- Adds infrastructure for a new lambda and API Gateway path
- Modifies build/deployment steps to support multiple lambdas based on the same .zip, and to include required dependencies
- A bit of re-factoring which is probably going to cause conflicts with @alexduf's work - sorry! 🙈 

In the next PR this lambda will start reading the access tokens generated by (https://github.com/guardian/mobile-purchases/pull/42) from S3. (We use an environment variable for now just to keep local testing simple).

I'll also add proper JSON responses based on the subscription status in the next PR (as you can see we are just using logging / generic placeholder responses for now to get something simple up and running).


google/playsubstatus/playsubstatus.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const access_token = process.env.AccessToken;
+
+interface GoogleResponseBody {
+    expiryTimeMillis: string
+}
+
+export async function handler(request: HTTPRequest): Promise<HTTPResponse> {
+
+    if (request.pathParameters && request.headers && request.headers["Play-Purchase-Token"]) {
```
alexduf: I know that semantically what you're doing is a GET, but I find it strange to have to use headers here to pass all the parameters.
Should that be modified to be a post with a proper json payload?

build-tc.sh
```diff
@@ -19,5 +19,5 @@ pushd tsc-target
 yarn install --production
 popd
 
-cd tsc-target/google
-zip ../mobile-purchases-google.zip *.js
\ No newline at end of file
+cd tsc-target
+zip -r mobile-purchases-google.zip . -i *.js
```
jacobwinch: We now need to include `.js` files from multiple subfolders (we also need to pull dependencies into the `.zip`)

google/playsubstatus/debug.ts
```diff
@@ -0,0 +1,12 @@
+import { handler } from './playsubstatus';
+
+let result = handler({
+    headers: {
+        "Play-Purchase-Token": "hoofgeonpgdfcnjdiofeadbe.AO-J1Oz1chlwrlUQnFQDBD2J3JDFiVPXnvkG7BYSYrv9FsjdWf4yyJd76jDhorOcTUU-Zx3r2VYccNAlFGljcdvPnsglwXWivMrR9b6rE7OIg4WGdLCR1W9bFu_lQvltMnQUQlGgkZ4WYe_eTt44wml5TatT70ew8g"
```
jacobwinch: This is a test token that I generated for myself

riff-raff.yaml
```diff
@@ -25,6 +25,17 @@ deployments:
       fileName: mobile-purchases-iosuserpurchases.jar
       prefixStack: false
 
+  mobile-purchases-google:
```
jacobwinch: Sorry, I ended up moving this whole block when resolving conflicts, but all we're really doing here is renaming the deployment task and updating a second lambda function with the same `.zip`

alexduf: yep sounds good!

google/playsubstatus/playsubstatus.ts
jacobwinch: Ha, funny you should say that... that's the way I originally implemented this (https://github.com/guardian/mobile-purchases/pull/43/commits/8dcbb7873cace0f6c0b3987136d7fab3f5a5ef0a).

I also had a working version which passed all of the details as path parameters, but I changed my mind about that this morning too (https://github.com/guardian/mobile-purchases/pull/43/commits/fe55afac7bc2e361613f18a5c5e20ed7d30ccc68), because I'd prefer to keep the purchase token out of the URL.

Overall, I felt it was stranger to use a `POST` than to set custom headers, but I'm not especially opinionated about this if you prefer the former approach?

PR #46, opened 2019-05-30
Parse the google play store payload and print it

alexduf:
Deployed and tested on CODE ✔️

### DIFF
--- a/google/src/models/apiGatewayHttp.ts
+++ b/google/src/models/apiGatewayHttp.ts
@@ -1,13 +1,13 @@
 export interface QueryParameters {
-    secret: string
+    [key: string]: string
 }
 
 export interface PathParameters {
-    subscriptionId: string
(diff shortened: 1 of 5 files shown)


google/src/models/apiGatewayHttp.ts
```diff
@@ -1,13 +1,13 @@
 export interface QueryParameters {
-    secret: string
+    [key: string]: string
```
alexduf: @jacobwinch let me know what you think of that change. It's handy but we lose typesafety.
That being said, if the parameter isn't set, no matter how typesafe the language is, we'll get an error.

jacobwinch: I think it's worth it overall... otherwise we are going to end up with loads of optional keys which are irrelevant for some of the lambdas anyway

PR #47, opened 2019-05-30
Write Google Access Tokens to S3

jacobwinch:
Tested and deployed to CODE ✅ 

Also fixes the packaging for the TypeScript lambdas so that dependencies are included in the `.zip`.


build-tc.sh
```diff
@@ -19,5 +19,5 @@ pushd tsc-target
 yarn install --production
 popd
 
-cd tsc-target/google/src
```
jacobwinch: If we `cd` to this directory then the `zip` command doesn't pull in the dependencies (I noticed that this was broken as I happened to test `mobile-purchases-googleplaysubstatus-CODE`).

alexduf: My bad!

But know we're also bundling the test source with it. This shouldn't block your PR though, so I'll make a note to fix that later

jacobwinch: Thanks - and good point 👍

PR #48, opened 2019-05-31
Read access tokens from S3 and improve build script

jacobwinch:
Now that the access tokens are [stored in S3](https://github.com/guardian/mobile-purchases/pull/47), we can start using them properly!

I've also made some more changes to the build script, as I ran into problems with missing modules when testing this change initially.

### DIFF
--- a/build-tc.sh
+++ b/build-tc.sh
@@ -6,18 +6,20 @@ npm install -g yarn
 mkdir -p tsc-target
 
 yarn install
-# Will place .js files in target
 yarn run clean
+
+# Will place .js files in tsc-target
 yarn run build
 
 yarn run test
 
 cp package.json tsc-target/
 
-pushd tsc-target
-# Ensures the RiffRaff package has the node_modules needed to run
-yarn install --production
-popd
-
 cd tsc-target
-zip -r mobile-purchases-google.zip . -i *.js
+
+# Create the .zip which will be uploaded to AWS
+zip -r mobile-purchases-google.zip . -i /google/src/*.js
+
+# Ensure node_modules (which are required by the lambdas at runtime) are included in the .zip
+yarn install --production
+zip -u -r mobile-purchases-google.zip node_modules/*
--- a/google/src/playsubstatus/playsubstatus.ts
+++ b/google/src/playsubstatus/playsubstatus.ts
@@ -1,38 +1,78 @@
 import * as restm from 'typed-rest-client/RestClient';
-import {HTTPResponseHeaders, HTTPRequest, HTTPResponse} from '../models/apiGatewayHttp';
-
-const access_token = process.env.AccessToken;
+import {HTTPResponseHeaders, HTTPRequest, HTTPResponse, HTTPResponses} from '../models/apiGatewayHttp';
+import S3 = require("aws-sdk/clients/s3");
 
 interface GoogleResponseBody {
     expiryTimeMillis: string
 }
 
+interface AccessToken {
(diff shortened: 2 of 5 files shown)


build-tc.sh
```diff
@@ -6,18 +6,20 @@ npm install -g yarn
...
-# Ensures the RiffRaff package has the node_modules needed to run
-yarn install --production
-popd
-
 cd tsc-target
-zip -r mobile-purchases-google.zip . -i *.js
\ No newline at end of file
+
+# Create the .zip which will be uploaded to AWS
+zip -r mobile-purchases-google.zip . -i /google/src/*.js
```
jacobwinch: This should address: https://github.com/guardian/mobile-purchases/pull/47#discussion_r289040544

build-tc.sh
```diff
@@ -6,18 +6,20 @@ npm install -g yarn
...
+
+# Ensure node_modules (which are required by the lambdas at runtime) are included in the .zip
+yarn install --production
+zip -u -r mobile-purchases-google.zip node_modules/*
```
jacobwinch: Perhaps we don't need absolutely everything in this folder, but I don't understand it well enough to figure out what can be safely excluded and I have been hitting issues at runtime.

google/tests/playsubstatus/playsubstatus.test.ts
```diff
@@ -1,15 +1,17 @@
...
+    // let handlerResult = handler({
+    //     headers: {
+    //         "Play-Purchase-Token": "hoofgeonpgdfcnjdiofeadbe.AO-J1Oz1chlwrlUQnFQDBD2J3JDFiVPXnvkG7BYSYrv9FsjdWf4yyJd76jDhorOcTUU-Zx3r2VYccNAlFGljcdvPnsglwXWivMrR9b6rE7OIg4WGdLCR1W9bFu_lQvltMnQUQlGgkZ4WYe_eTt44wml5TatT70ew8g"
+    //     },
+    //     pathParameters: {
+    //         "subscriptionId": "com.guardian.subscription.monthly.10"
+    //     }
+    // });
+    //
+    // return handlerResult.then(result => expect(result.statusCode).toBe(200));
```
jacobwinch: This test won't fail silently anymore (fixed as per https://github.com/guardian/mobile-purchases/pull/45), but I've commented it out again for now because it won't work when run via TeamCity

google/src/playsubstatus/playsubstatus.ts
```diff
@@ -1,38 +1,78 @@
...
+const s3: S3 = new S3();
+
+function getParams(stage: string): S3.Types.GetObjectRequest {
+    return {
+        Bucket: 'gu-mobile-access-tokens',
+        Key: `${stage}/google-play-developer-api/access_token.json`
+    }
+}
+
+function getAccessToken(params: S3.Types.GetObjectRequest): Promise<AccessToken> {
```
jacobwinch: As a future optimisation, we should figure out how we can allow AWS to cache this safely (as discussed with @alexduf)

build-tc.sh
alexduf: As chatted this will be superseded by Webpack

google/src/playsubstatus/playsubstatus.ts
alexduf: sounds good. It doesn't need to be done in this PR, but basically you can define a global state. Happy to pair on that subject

PR #49, opened 2019-05-31
Add alert for Google token refresh failures

jacobwinch:
A number of services in this project (will) depend on us successfully refreshing access tokens for the Google Play Developer API. Therefore we should be informed if there are any problems refreshing this token in PROD.

Since this lambda [throws an exception](https://github.com/guardian/mobile-purchases/blob/master/googleoauth/src/main/scala/com.gu.mobilepurchases.googleoauth/lambda/GoogleOAuth.scala#L30-L33) on error, we should be able to use standard `AWS/Lambda` metrics to detect token refresh failures.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -382,7 +382,8 @@ Resources:
     Properties:
       AlarmActions:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:jacob-test # FIXME should be mobile-server-side
-      AlarmName: !Sub Mobile Purchases ${Stage} | Failed to refresh Google OAuth Token
+      AlarmName: !Sub mobile-purchases-${Stage}-google-oauth-token-refresh-failure
+      AlarmDescription: !Sub Trigger the GoogleOAuth lambda manually to refresh the token
       ComparisonOperator: GreaterThanOrEqualToThreshold
       Dimensions:
         - Name: FunctionName
@@ -390,7 +391,7 @@ Resources:
       EvaluationPeriods: 1
       MetricName: Errors
       Namespace: AWS/Lambda
-      Period: 300
+      Period: 60
       Statistic: Sum
       Threshold: 1
       TreatMissingData: notBreaching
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -373,3 +375,23 @@ Resources:
         Stage: !Ref Stage
         Stack: !Ref Stack
         App: !Ref App
+
+  GoogleTokenRefreshFailureAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdOnlyResources
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:mobile-server-side
```
jacobwinch: I wasn't entirely sure which topic to use, so please let me know if there's a better choice

cloudformation.yaml
```diff
@@ -373,3 +375,23 @@ Resources:
...
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Dimensions:
+        - Name: FunctionName
+          Value: !Ref GoogleOAuthLambda
+      EvaluationPeriods: 1
+      MetricName: Errors
+      Namespace: AWS/Lambda
+      Period: 60
+      Statistic: Sum
+      Threshold: 1
```
jacobwinch: This might prove to be oversensitive in the long term, but I prefer to remain cautious when introducing a new service...

cloudformation.yaml
alexduf: Nope that's the right one :)

cloudformation.yaml
alexduf: Yeah let's see if it's too verbose. Better safe than sorry!

cloudformation.yaml
```diff
@@ -35,6 +35,8 @@ Parameters:
   GooglePubSubSecret:
     Type: String
     Description: The secret used by google's pubsub
+Conditions:
```
alexduf: I would create it in CODE too!

PR #50, opened 2019-05-31
Store event dynamo and introduce webpack

alexduf:
Start storing events in the dynamo DB table.

Also I've introduced webpack which means each function has its own bundle now, that makes the packaging of dependencies much straight forward


alexduf:
Thanks!

PR #52, opened 2019-05-31
Send subscriptionId messages to a queue

alexduf:
This creates an SQS queue via cloudformation, then posts subscriptionIds to it.

For the moment nothing reads that queue.


cloudformation.yaml
```diff
@@ -376,6 +384,18 @@ Resources:
         Stack: !Ref Stack
         App: !Ref App
 
+  GoogleSubscriptionsQueue:
```
jacobwinch: We might want to [encrypt](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#aws-sqs-queue-kmsmasterkeyid) this queue? 

Perhaps we should also Cloudform a dead letters queue?

alexduf: That's a good shout on both these points!

alexduf:
I haven't yet managed to prove this works on CODE, so I won't merge quite yet :)

PR #53, opened 2019-06-03
Re-organise project

alexduf:
Split the two technologies (scala and typescript) into two different folders.

As I was about to add an `Apple` folder it started to get messy on the typescript side, so I took a step back and :

- split scala and typescript in two directory
- renamed scala modules to be separated-by-dashes ratherthanonebigword
- created a single typescript project rather than expecting to have one for the google side, one for the apple side as webpack should handle the dependency tree by itself. This will make code sharing easier between typescript lambdas


.gitignore
```diff
@@ -2,4 +2,5 @@
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 .idea
 target
-tsc-target
\ No newline at end of file
+tsc-target
+node_modules
```
jacobwinch: 💯

NathanielBennett:
:+1: Yup cool. Can you merge this before I start adding stuff to build.sbt?

alexduf:
Something went wrong when I pushed this branch. My bad, let me try to fix it!

PR #56, opened 2019-06-04
Add Apple pubsub support

alexduf:
This creates:
 - A new lambda to handle Apple's pubsub events
 - An encrypted SQS queue with its corresponding dead letter queue

It also adds support for proper stack traces (via npm package `source-map-support`) which makes debugging easier.

The PR is bigger than I usually like them to be as it required a significant rework of the google code, sorry!


alexduf:
Deployment requires to apply the cloudformation first as there's a new parameter

cloudformation.yaml
```diff
@@ -420,6 +467,34 @@ Resources:
...
+      KmsMasterKeyId: alias/aws/sqs
+      Tags:
+        - Key: Stage
+          Value: !Ref Stage
+        - Key: Stack
+          Value: !Ref Stack
+        - Key: App
+          Value: !Ref App
+
+  AppleSubscriptionsQueueDlq:
```
jacobwinch: Sorry to be a pedant but strictly speaking I think we probably need to encrypt this queue too?

alexduf: oh, you're absolutely right! that's not pedant at all

PR #57, opened 2019-06-05
Fix playsubstatus request validation

jacobwinch:
Headers fields are [case-insensitive](https://tools.ietf.org/html/rfc7230#section-3.2), which means that certain clients will sometimes lowercase the `Play-Purchase-Token` header. This PR allows the `playsubstatus` lambda to cope with this.

I noticed the OkHttp library doing this when testing via the Android App. @alexduf I wonder if this is also what was causing the mysterious issue when you were testing this with `curl`.

### DIFF
--- a/typescript/src/playsubstatus/playsubstatus.ts
+++ b/typescript/src/playsubstatus/playsubstatus.ts
@@ -1,5 +1,12 @@
 import * as restm from 'typed-rest-client/RestClient';
-import {HTTPResponseHeaders, HTTPRequest, HTTPResponse, HTTPResponses} from '../models/apiGatewayHttp';
+import {
+    HTTPResponseHeaders,
+    HTTPRequest,
+    HTTPResponse,
+    HTTPResponses,
+    HttpRequestHeaders,
+    PathParameters
+} from '../models/apiGatewayHttp';
 import S3 = require("aws-sdk/clients/s3");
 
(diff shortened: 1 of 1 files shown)


typescript/src/playsubstatus/playsubstatus.ts
```diff
@@ -41,13 +48,22 @@ function getAccessToken(params: S3.Types.GetObjectRequest): Promise<AccessToken>
         })
 }
 
+function getPurchaseToken(headers: HttpRequestHeaders): string {
+    return headers["Play-Purchase-Token"] || headers["play-purchase-token"]
+}
+
+function buildGoogleUrl(pathParams: PathParameters, headers: HttpRequestHeaders) {
+    const baseUrl = 'https://www.googleapis.com/androidpublisher/v3/applications/com.guardian/purchases/subscriptions';
+    return `${baseUrl}/${pathParams.subscriptionId}/tokens/${getPurchaseToken(headers)}`;
```
alexduf: Is that a javascript thing or a typescript thing? I like it very much!

jacobwinch: I think you mean this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals. In which case, yep I agree it's pretty neat!

PR #62, opened 2019-06-11
Handle 410 response from Google

jacobwinch:
Prior to this PR, we were serving a 500 when attempting to look up subscriptions which expired a long time ago.

See https://stackoverflow.com/questions/45688494/google-android-publisher-api-responds-with-410-purchasetokennolongervalid-erro for more details.

### DIFF
--- a/typescript/src/playsubstatus/playsubstatus.ts
+++ b/typescript/src/playsubstatus/playsubstatus.ts
@@ -83,7 +83,7 @@ export async function handler(request: HTTPRequest): Promise<HTTPResponse> {
             })
             .catch(
                 error => {
-                    if (error['statusCode'] === 410) {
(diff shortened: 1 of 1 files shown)


typescript/src/playsubstatus/playsubstatus.ts
```diff
@@ -82,15 +82,18 @@ export async function handler(request: HTTPRequest): Promise<HTTPResponse> {
                 }
             })
             .catch(
-                error =>  {
-                    console.log(`Serving an Internal Server Error due to: ${error}`);
-                    return HTTPResponses.INTERNAL_ERROR
+                error => {
+                    if (error['statusCode'] === 410) {
```
alexduf: Why not using?
```suggestion
                    if (error.statusCode === 410) {
```

PR #65, opened 2019-06-17
User link logic

alexduf:
This only provides the parsing logic as I have to change priority and would like to ship what I had up to now

### DIFF
--- a/typescript/src/utils/receiptParser.ts
+++ b/typescript/src/utils/receiptParser.ts
@@ -1,0 +1,39 @@
+import {AppleReceiptInfo} from "../models/appleReceiptInfo";
+
+interface DecodedReceipt {
+    signature: string,
+    purchase_info: string,
+    pod: string,
+    signing_status: string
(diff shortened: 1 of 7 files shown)


typescript/src/utils/receiptParser.ts
```diff
@@ -0,0 +1,39 @@
...
+    purchase_info: string,
+    pod: string,
+    signing_status: string
+}
+
+function normaliseKeyName(key: string): string {
+    return key.replace(/-/g, "_")
+}
+
+export function parseStrangeAppleJson(strangeJson: string): any {
```
jacobwinch: 😂

alexduf: It's true, it's JSON without being JSON. Apple trying to be a cool kid by not re-using any popular technology.

Whatever format that is I couldn't find a name, nor a lightweight library to parse it. So I think Apple fully deserves this method name

PR #66, opened 2019-06-24
Reconfigure teamcity build to run on docker container

AWare:
This will run in a docker container which doesn't have zip, but does have yarn.
`build-tc-2.sh` will run outside the container (this might not want to be a file, but just a separate build step)

Will make teamcity changes to accompany merging.


alexduf:
closing in favour of 
https://github.com/guardian/mobile-purchases/pull/72

PR #67, opened 2019-06-26
Mss 901/update subscriptions lambda

NathanielBennett:
A new lambda, triggered by the Google subscription queue, consumes incoming messages and persists subscriptions to to dynamo


typescript/src/models/subscription.ts
```diff
@@ -0,0 +1,41 @@
...
+        this.startTimeStamp = startTimeStamp;
+        this.endTimeStamp = endTimeStamp;
+        this.cancellationTimetamp = cancellationTimeStamp;
+        this.autoRenewing = autoRenewing;
+        this.googlePayload = googlePayload;
+        this.ttl = ttl;
+    }
+
+    get [DynamoDbTable]() {
+        return App + "-" + Stage + "-subscriptions"
```
webb04: can also do: \`${App}-${Stage}-subscriptions\`

typescript/src/utils/aws.ts
```diff
@@ -0,0 +1,34 @@
...
+    function () { return new ECSCredentials(); },
+    function () { return new SharedIniFileCredentials({
+        profile: "mobile"
+    }); }
+]);
+
+const aws = new DynamoDB({
+    region: Region,
+    credentialProvider: credentialProvider
+});
```
webb04: You can also do this in JavaScript
```
const aws = new DynamoDB({
    region: Region,
    credentialProvider
});
```

typescript/src/updatesubs/google.ts
```diff
@@ -0,0 +1,69 @@
+process.on('uncaughtException', function (err) { console.log(err); })
```
webb04: Do we want this for logging?

typescript/src/utils/google-play.ts
```diff
@@ -0,0 +1,36 @@
+import aws = require("../utils/aws");
```
webb04: Could you just import s3? Ah you have the s3 type below, so maybe this is clearer.

typescript/src/utils/aws.ts
NathanielBennett: I don't think it looks as clean though

typescript/src/updatesubs/google.ts
NathanielBennett: No I need to get rid of it

PR #74, opened 2019-07-18
Mss 954/user subscription lambda

NathanielBennett:
Implements the user lambda as outlined here: https://theguardian.atlassian.net/secure/RapidBoard.jspa?rapidView=18&projectKey=MSS&modal=detail&selectedIssue=MSS-954. 

Note: This changes the userSubscription table to be keyed on userId instead of subscriptionId. This is so all subscriptions for a user can be retrieved

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -238,7 +238,7 @@ Resources:
           "/subscriptions":
             get:
               x-amazon-apigateway-integration:
-                httpMethod: GET
+                httpMethod: POST
                 type: aws_proxy
                 uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UserSubscriptionsLambda.Arn}/invocations
               consumes: [ application/json ]
@@ -714,4 +714,3 @@ Resources:
 
 
 
-      
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -234,6 +234,18 @@ Resources:
                 "200":
                   "description": "200 response"
 
+          "/subscriptions":
+            post:
+              x-amazon-apigateway-integration:
+                httpMethod: POST
+                type: aws_proxy
+                uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UserSubscriptionsLambda.Arn}/invocations
+              consumes: [ application/json ]
+              produces: [ application/json ]
+              responses:
+                "200":
+                  "description": "200 response"
+
           "/healthcheck":
             get:
               responses:
@@ -668,6 +680,35 @@ Resources:
       BatchSize: 10
 
 
+  UserSubscriptionsLambda:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: user-subscriptions.handler
+      Runtime: nodejs10.x
+      CodeUri:
+        Bucket: !Ref DeployBucket
+        Key: !Sub ${Stack}/${Stage}/${App}-user-subscriptions/user-subscriptions.zip
+      FunctionName: !Sub ${App}-user-subscriptions-${Stage}
+      Role: !GetAtt MobilePurchasesLambdasRole.Arn
+      Environment:
+        Variables:
+          Stage: !Ref Stage
+          Stack: !Ref Stack
+          App: !Ref App
+      Description: Retrieves subscription details for a given user
+      MemorySize: 128
+      Timeout: 45
+      Events:
+        GetApi:
+          Type: Api
+          Properties:
+            Path: "/subscriptions"
+            Method: GET
+            RestApiId: !Ref MobilePuchasesApi
+      Tags:
+        Stage: !Ref Stage
+        Stack: !Ref Stack
+        App: !Ref App
 
 
 
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -103,6 +103,7 @@ Resources:
               Effect: Allow
               Action:
                   - "dynamodb:GetItem"
+                  - "dynamodb:BatchGetItem"
                   - "dynamodb:PutItem"
                   - "dynamodb:UpdateItem"
                   - "dynamodb:Query"
@@ -235,7 +236,7 @@ Resources:
                   "description": "200 response"
 
           "/subscriptions":
-            post:
+            get:
               x-amazon-apigateway-integration:
                 httpMethod: POST
                 type: aws_proxy
@@ -699,7 +700,7 @@ Resources:
       MemorySize: 128
       Timeout: 45
       Events:
-        GetApi:
+        PostApi:
           Type: Api
           Properties:
             Path: "/subscriptions"
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -238,7 +238,7 @@ Resources:
           "/subscriptions":
             get:
               x-amazon-apigateway-integration:
-                httpMethod: POST
+                httpMethod: GET
                 type: aws_proxy
                 uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UserSubscriptionsLambda.Arn}/invocations
               consumes: [ application/json ]
--- a/dynamo.cloudformation.yaml
+++ b/dynamo.cloudformation.yaml
@@ -113,9 +113,9 @@ Resources:
         - AttributeName: userId
           AttributeType: S
       KeySchema:
-        - AttributeName: subscriptionId
-          KeyType: HASH
         - AttributeName: userId
+          KeyType: HASH
+        - AttributeName: subscriptionId
           KeyType: RANGE
       PointInTimeRecoverySpecification:
         PointInTimeRecoveryEnabled: true
--- a/typescript/src/models/subscription.ts
+++ b/typescript/src/models/subscription.ts
@@ -39,3 +39,32 @@ export class Subscription {
         return `${App}-${Stage}-subscriptions`
     }
 }
+
+export class ReadSubscription {
+    @hashKey()
+    subscriptionId: string;
+
+    @attribute()
+    startTimeStamp: string;
+
+    @attribute()
+    endTimeStamp: string;
+
+    @attribute()
+    cancellationTimetamp?: string;
+
+    @attribute()
+    autoRenewing?: boolean;
+
+    constructor() {
+        this.subscriptionId = "";
+        this.startTimeStamp = "";
+        this.endTimeStamp = "";
+        this.cancellationTimetamp = ""
+        this.autoRenewing = false;
+    }
+
+    get [DynamoDbTable]() {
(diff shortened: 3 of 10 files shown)


typescript/src/user/user.ts
```diff
@@ -0,0 +1,112 @@
...
+import {tagFilter} from "aws-sdk/clients/health";
+import {subscriptionARN} from "aws-sdk/clients/sns";
+import {getUserId, getIdentityToken} from "../utils/guIdentityApi";
+
+class SubscriptionStatus {
+   subscriptionId: string;
+   from?: string;
+   to?: string;
+   status: string;
+   cancelled?: string;
```
davidfurey: I would expect `cancelled` to be a boolean.  It looks like this is actually `cancellationTimetamp`, why not call it that?

typescript/src/user/user.ts
```diff
@@ -0,0 +1,112 @@
...
+      this.subscriptionId = subscription.subscriptionId;
+      this.from = subscription.startTimeStamp;
+      this.to = subscription.endTimeStamp;
+      this.status = this.getStatus(subscription.endTimeStamp, subscription.cancellationTimetamp);
+      this.cancelled = subscription.cancellationTimetamp === "" ? undefined : subscription.cancellationTimetamp
+   }
+}
+
+class SubscriptionStatusResponse {
+    activeSubscriptions: string[];
```
davidfurey: Maybe `activeSubscriptionIds`?

typescript/src/user/user.ts
```diff
@@ -0,0 +1,112 @@
...
+async function getSubscriptions(subscriptionIds: string[]) : Promise<SubscriptionStatusResponse>  {
+    const subs: ReadSubscription[] = [];
+    const toGet = subscriptionIds.map( subscriptionId => Object.assign( new ReadSubscription, {subscriptionId: subscriptionId} ) )
+
+    for await (const sub of dynamoMapper.batchGet(toGet)  ) {
+        subs.push(sub)
+    }
+
+    const sortedSubs = subs.sort((subscriptionA: Subscription, subscriptionB: Subscription) => {
+        const endTimeA = subscriptionA.endTimeStamp && Date.parse(subscriptionA.endTimeStamp) || 0
```
davidfurey: It looks like you are concerned about endTimeStamp being null here, but not on line 53, why?

typescript/src/user/user.ts
```diff
@@ -0,0 +1,112 @@
...
+        return endTimeA - endTimeB
+    })
+    return new SubscriptionStatusResponse(sortedSubs)
+}
+
+
+export async function handler(httpRequest: HTTPRequest): Promise<HTTPResponse> {
+    if(httpRequest.headers && getIdentityToken(httpRequest.headers)) {
+        return getUserId(httpRequest.headers)
+            .then( userId => {
```
davidfurey: Do you need the `return` here or could this be more concisely written as
`.then( userId => getUserSubscriptionIds(userId))`

typescript/src/utils/guIdentityApi.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function getIdentityToken(headers: HttpRequestHeaders): string {
+    return headers["Gu-Identity-Token"] || headers["gu-identity-token"]
+}
+
+export function getUserId(headers: HttpRequestHeaders): Promise<string> {
+    const url = "https://id.guardianapis.com/user/me"
+    const identityToken = getIdentityToken(headers)
+    console.log(`Identity token: *${identityToken}*`)
```
davidfurey: Is this safe to log?

typescript/src/user/user.ts
NathanielBennett: It's because timestamp is an optional value in the subscription model class. This is so we can doi a database lookkup without needing to know the timestamp values. (I didn't decide to use dbmapper ). So here, i need to treat the endtimes as potentially undefined in order to do a compare them.

typescript/src/user/user.ts
NathanielBennett: It's a similar issue to what we saw when we were pairing where without the explicit return, the lamba does wait for the future to complete

typescript/src/utils/guIdentityApi.ts
NathanielBennett: It's probably safe, but unneccesary, so removing

typescript/src/user/user.ts
davidfurey: Strange.  JS doesn't work how I expect it to.

cloudformation.yaml
```diff
@@ -234,6 +235,18 @@ Resources:
                 "200":
                   "description": "200 response"
 
+          "/subscriptions":
+            get:
+              x-amazon-apigateway-integration:
+                httpMethod: POST
```
alexduf: GET

alexduf: GET

dynamo.cloudformation.yaml
```diff
@@ -113,9 +113,9 @@ Resources:
         - AttributeName: userId
           AttributeType: S
       KeySchema:
-        - AttributeName: subscriptionId
-          KeyType: HASH
         - AttributeName: userId
+          KeyType: HASH
+        - AttributeName: subscriptionId
```
alexduf: Nice one!

typescript/src/models/subscription.ts
```diff
@@ -35,6 +35,35 @@ export class Subscription {
         this.ttl = ttl;
     }
 
+    get [DynamoDbTable]() {
+        return `${App}-${Stage}-subscriptions`
+    }
+}
+
+export class ReadSubscription {
```
alexduf: What's the difference between the Subscription and the ReadSubscription, I'm not sure I understand why it's defined twice

typescript/src/user/user.ts
```diff
@@ -0,0 +1,112 @@
...
+import {SubscriptionUpdate} from "../updatesubs/updatesub";
+import {tagFilter} from "aws-sdk/clients/health";
+import {subscriptionARN} from "aws-sdk/clients/sns";
+import {getUserId, getIdentityToken} from "../utils/guIdentityApi";
+
+class SubscriptionStatus {
+   subscriptionId: string;
+   from?: string;
+   to?: string;
+   status: string;
```
alexduf: It would be nice to define a type for the status.

Something like

```typescript
type SubscriptionStatusEnum = "active" | "expired" | "wontRenew";
```

typescript/src/user/user.ts
```diff
@@ -0,0 +1,112 @@
...
+
+   private getStatus(endTimeStamp: string, cancellelationTimestamp?: string) {
+
+       const now = Date.now()
+
+       if (cancellelationTimestamp === "" || cancellelationTimestamp === undefined)   {
+          return Date.parse(endTimeStamp) > now ? "active" : "expired"
+       }
+       else {
+           return "cancellationTimestamp"
```
alexduf: I'm thinking `"wontRenew"` or `"expireSoon"`

typescript/src/user/user.ts
alexduf: I think the rule is, if it's a one line with one statement you can write it within brackets and without a return. If you need more, then use curly braces but you need a return statement.

so this:
```typescript
.then( userId => getUserSubscriptionIds(userId))
```

is the same as this

```typescript
.then( userId => {
  return getUserSubscriptionIds(userId)
})

File: typescript/src/user/user.ts (hunk 10 above)
Comment (alexduf) [reply typescript/src/user/user.ts]: Maybe it needs to be broken down in two flags.
One that conveys whether it's valid or not, one that conveys whether the subscription will renew or not

File: typescript/src/models/subscription.ts (hunk 8 above)
Comment (NathanielBennett) [reply typescript/src/models/subscription.ts]: It's because I can't do a batch get of subscription, without a zero arg constructor - and you can't have multiple constructors. See here: https://github.com/awslabs/dynamodb-data-mapper-js

File: typescript/src/user/user.ts (hunk 10 above)
Comment (NathanielBennett) [reply typescript/src/user/user.ts]: There's no cancellation timestamp unless the subscription has actually been cancelled. We've got an ''autoRenewing' boolean in the subscription table. We could add that to the response?

File: typescript/src/user/user.ts (hunk 4 above)
Comment (NathanielBennett) [reply typescript/src/user/user.ts]: tsc doesn't like that. I assume because of the return statement: 

```typescript/src/user/user.ts:95:30 - error TS1005: '{' expected.
95             .then( userId => return getUserSubscriptionIds(userId); )
                                ~~~~~~
typescript/src/user/user.ts:95:69 - error TS1128: Declaration or statement expected.
95             .then( userId => return getUserSubscriptionIds(userId); )```
                                                                       ~

File: dynamo.cloudformation.yaml (hunk 7 above)
Comment (NathanielBennett) [reply dynamo.cloudformation.yaml]: Thanks!

File: typescript/src/user/user.ts (hunk 4 above)
Comment (NathanielBennett) [reply typescript/src/user/user.ts]: I found that i explicitly needed the return to make aws wait for the promise ..
Sorry :(

File: typescript/src/models/subscription.ts (hunk 8 above)
Comment (alexduf) [reply typescript/src/models/subscription.ts]: I see!
Are you sure you need a batch get? I think a scan is what you actually want!

File: typescript/src/user/user.ts (hunk 10 above)
Comment (alexduf) [reply typescript/src/user/user.ts]: Yes, because as soon as the autorenewing flag goes to false we'll have the cancellation date.

File: typescript/src/user/user.ts (hunk 4 above)
Comment (alexduf) [reply typescript/src/user/user.ts]: In your code you have the return statement that is not needed:  
`.then( userId => return getUserSubscriptionIds(userId); )` that can be written
`.then( userId => getUserSubscriptionIds(userId))`

You may have had a problem in the past if you wrote it with curly braces like that:
` .then( userId => {getUserSubscriptionIds(userId)})` => that won't return what you want

Comment (alexduf) [reply typescript/src/user/user.ts]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Shorter_functions

File: typescript/src/models/subscription.ts (hunk 8 above)
Comment (NathanielBennett) [reply typescript/src/models/subscription.ts]: I don't think a scan will work here: That would be the case if there were multiple records with the same subscriptionId. But here we want multiple items with different subscriptionIds!

File: typescript/src/user/user.ts (hunk 4 above)
Comment (NathanielBennett) [reply typescript/src/user/user.ts]:
```            .then( userId => getUserSubscriptionIds(userId))
            .then(subIds => getSubscriptions(subIds))
            .then( subs => {
                return new HTTPResponse(200, new HTTPResponseHeaders(), JSON.stringify(subs) )
            })
```
I'm going to try this bu I'm not conviced that aws won't exit the lambda before the promises return

typescript/src/models/subscription.ts
alexduf: We want to scan all the record with the same user ID right?

PR #75, opened 2019-07-23
Mss 952/apple update subscription lambda

NathanielBennett:
Implements storage of apple subscriptions: Needs https://github.com/guardian/mobile-purchases/pull/74 to be merged first ( this is where the dynamo cfn changes come from )

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -111,7 +111,7 @@ Resources:
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-${Stack}-user-purchases
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscription-events
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscriptions
-                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
+                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-TEST-subscriptions
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
 
 
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -111,7 +111,6 @@ Resources:
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-${Stack}-user-purchases
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscription-events
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscriptions
-                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-TEST-subscriptions
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
 
 
@@ -123,7 +122,6 @@ Resources:
               Resource:
                 - !GetAtt GoogleSubscriptionsQueue.Arn
                 - !GetAtt AppleSubscriptionsQueue.Arn
-                - arn:aws:sqs:eu-west-1:201359054765:NathanielUpdateGoogleSubscriptionTst
         - PolicyName: Kms
           PolicyDocument:
             Statement:
@@ -738,7 +736,7 @@ Resources:
     Properties:
       FunctionName: !Ref UpdateAppleSubscriptionsLambda
       Enabled: true
-      EventSourceArn: arn:aws:sqs:eu-west-1:201359054765:NathanielUpdateGoogleSubscriptionTst
+      EventSourceArn: !GetAtt AppleSubscriptionsQueue.Arn
       BatchSize: 10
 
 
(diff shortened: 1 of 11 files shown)


typescript/src/updatesubs/apple.ts
```diff
@@ -0,0 +1,46 @@
...
+        await parseAndStoreSubscriptionUpdate(record, makeDynamoPromise)
+    })
+
+    return Promise.all(emptyPromises)
+        .then( value => {
+            console.log(`Processed ${event.Records.length} subscriptions`)
+            "OL"
+        })
+        .catch(error => {
+            console.error("Error processing subsctption update: ", error)
```
davidfurey: Typo "subsctption"

typescript/src/updatesubs/apple.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function handler(event: SQSEvent) {
+    const emptyPromises = event.Records.map( async (record) => {
+        await parseAndStoreSubscriptionUpdate(record, makeDynamoPromise)
+    })
+
+    return Promise.all(emptyPromises)
+        .then( value => {
+            console.log(`Processed ${event.Records.length} subscriptions`)
+            "OL"
```
davidfurey: What does "OL" mean?

typescript/src/updatesubs/google.ts
```diff
@@ -22,7 +26,23 @@ interface GoogleResponseBody {
 
 const restClient = new restm.RestClient('guardian-mobile-purchases');
 
-export function getGoogleSubResponse(record: SQSRecord): Promise<SubscriptionUpdate> {
+function parseAutoRenewing(autoRenewing: string | undefined) : boolean | undefined {
+
+        if(autoRenewing) {
+            try {
+                return JSON.parse(autoRenewing)
```
davidfurey: Seems like overkill to use JSON.parse to check if a string is true or false

typescript/src/updatesubs/updatesub.ts
```diff
@@ -1,18 +1,18 @@
...
     if (cancellationTime) {
         return new Date(Number.parseInt(cancellationTime)).toISOString()
     } else {
         return ""
     }
 }
 
-function makeTimeToLive(date: Date) {
+export function makeTimeToLive(date: Date) {
     return Math.ceil((date.getTime() / 1000) + 7 * ONE_YEAR_IN_SECONDS)
```
davidfurey: Maybe not one for this PR, but I don't think we should be keeping this data for 7 years

typescript/src/models/subscription.ts
```diff
@@ -19,23 +20,49 @@ export class Subscription {
...
-    googlePayload?: any;
-
     @attribute()
     ttl?: number;
 
-    constructor(subscriptionId: string, startTimeStamp?: string, endTimeStamp?: string, cancellationTimeStamp?: string, autoRenewing?: boolean, googlePayload?: any,  ttl?: number) {
+    constructor(subscriptionId: string, startTimeStamp?: string, endTimeStamp?: string, cancellationTimeStamp?: string, autoRenewing?: boolean, ttl?: number) {
         this.subscriptionId = subscriptionId;
-        this.startTimeStamp = startTimeStamp;
+        this.startTimeStamp = startTimeStamp
```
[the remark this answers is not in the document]
NathanielBennett: Good point. Not yet. Although ts semis are a bit like scala ones, right?

typescript/src/updatesubs/google.ts
NathanielBennett: Actually it doesn't really matter because It's not actually used. Binning it.

typescript/src/updatesubs/updatesub.ts
NathanielBennett: OK, that was the original requirement. See @alexduf doxs.

cloudformation.yaml
```diff
@@ -668,6 +668,35 @@ Resources:
...
+      FunctionName: !Sub ${App}-apple-update-subscriptions-${Stage}
+      Code:
+        S3Bucket: !Ref DeployBucket
+        S3Key: !Sub ${Stack}/${Stage}/${App}-apple-update-subscriptions/apple-update-subscriptions.zip
+      Environment:
+        Variables:
+          App: !Sub ${App}
+          Stack: !Sub ${Stack}
+          Stage: !Sub ${Stage}
+      Description: Consomes subscription data updates from app store from sqs and stores them in dynamo
```
alexduf:
```suggestion
      Description: Consumes subscription data updates from app store from sqs and stores them in dynamo
```

typescript/src/updatesubs/apple.ts
```diff
@@ -0,0 +1,46 @@
...
+    receipt: string,
+    transactionId: string,
+    startDate: string,
+    cancellationDate: string,
+    endDate: string,
+    autoRenewing: boolean
+    appleBody: string
+}
+
+export function makeDynamoPromise(record: SQSRecord): Promise<AppleSubscription>  {
```
alexduf: I think this method it misnamed, as currently it converts  a model to another model.

Now, I think we need to either verify the cryptographic signature that the receipt at the pubsub stage, or have an API call to Apple at this point to check the receipt is valid and up to date.

That might be a good case for a separate PR?

typescript/src/updatesubs/apple.ts
```diff
@@ -0,0 +1,46 @@
...
+        makeCancellationTime(sub.cancellationDate),
+        sub.autoRenewing,
+        sub.appleBody,
+        makeTimeToLive(new Date(Date.now()))
+    ))
+}
+
+export async function handler(event: SQSEvent) {
+    const emptyPromises = event.Records.map( async (record) => {
+        await parseAndStoreSubscriptionUpdate(record, makeDynamoPromise)
```
alexduf: here I think you need a return statement because of the curly braces

typescript/src/updatesubs/apple.ts
NathanielBennett: Agreed. Do we have details of he apple API call to make. Lets do this as a separate PR at the pubsub test

NathanielBennett:
I've created a story to validate the reciept: https://theguardian.atlassian.net/browse/MSS-1054

PR #79, opened 2019-08-16
Introduce a new error type to help with error handling.

alexduf:
There was no ticket for this work, I was reviewing the codebase as a whole before we release it.

The new error type carries a boolean flag to help the lambda handler decide whether or not to throw the exception or to swallow it. Some exceptions might be acceptable and the lambda considered a success, and some might require to be retried.

### DIFF
--- a/typescript/src/update-subs/apple.ts
+++ b/typescript/src/update-subs/apple.ts
@@ -7,6 +7,7 @@ import fetch from 'node-fetch';
 import {Response} from 'node-fetch';
 import {Stage} from "../utils/appIdentity";
 import {msToFormattedString, optionalMsToFormattedString} from "../utils/dates";
+import {ProcessingError} from "../models/processingError";
 
 // const receiptEndpoint = (Stage === "PROD") ? "https://buy.itunes.apple.com/verifyReceipt" : "https://sandbox.itunes.apple.com/verifyReceipt";
 // const environment = (Stage === "PROD") ? "Production" : "Sandbox";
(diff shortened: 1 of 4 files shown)


typescript/src/update-subs/apple.ts
```diff
@@ -38,21 +39,25 @@ function validateReceipt(subRef: AppleSubscriptionReference): Promise<Response>
...
+            throw new ProcessingError("Impossible to validate receipt", true);
         }
         return response
     })
 }
 
 function checkResponseStatus(response: AppleValidationResponse): AppleValidationResponse {
-    if (response.status != 0) {
+    if ((response.status >= 21100 && response.status <= 21199) || response["is-retryable"]) {
+        console.error(`Server error received from Apple, got status ${response.status} for ${response.latest_receipt}, will retry`);
```
jacobwinch: Is `latest_receipt` an id or an object? If it's the latter, is it acceptable to log the full receipt here (from a privacy perspective)?

alexduf: `latest_receipt` is an object. I guess in this case it doesn't make sense to log it as there's potentially nothing wrong with it. I'll remove it from here.

I think I'm okay with logging it if Apple replies it's malformed or invalid though

typescript/src/update-subs/updatesub.ts
```diff
@@ -40,12 +41,25 @@ function putSubscription(subscription: Subscription): Promise<Subscription>  {
...
+            return "OK"
+        })
+        .catch(error => {
+            if (error instanceof ProcessingError) {
+               console.error("Error processing the subscription update", error);
+               if (error.shouldRetry) {
+                   console.error("Will throw an exception to retry this message");
+                   throw error;
+               }  else {
+                   return "Error, giving up"
```
jacobwinch: Might be worth adding logging in this case too?

PR #80, opened 2019-08-16
Reduce rows TTL

alexduf:
This pull request:

- reduce the TTL for events to 30 months since the event
- reduce the TTL of subscription to 30 month after the expiry date of the subscription
- introduce a TTL for the userSubscription table

### DIFF
--- a/typescript/src/link/link.ts
+++ b/typescript/src/link/link.ts
@@ -6,6 +6,7 @@ import {dynamoMapper, sendToSqsImpl} from "../utils/aws";
 import {ItemNotFoundException} from "@aws/dynamodb-data-mapper";
 import {catchClause} from "@babel/types";
 import {getUserId, getIdentityToken} from "../utils/guIdentityApi";
+import {dateToSecondTimestamp, thirtyMonths} from "../utils/dates";
 
 export class UserSubscriptionData {
     transactionToken: string;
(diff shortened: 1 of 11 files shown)


typescript/src/link/link.ts
```diff
@@ -21,7 +22,8 @@ function putUserSubscription(subscriptionId: string, userId: string): Promise<Us
     const userSubscription = new UserSubscription(
         userId,
         subscriptionId,
-        new Date(Date.now()).toISOString()
+        new Date(Date.now()).toISOString(),
+        dateToSecondTimestamp(thirtyMonths())
```
jacobwinch: I wonder if this should also be 30 months _after subscription expiry date_... 

Because:

a) perhaps this consistency between tables would be helpful?
b) I can imagine a use case where we want to attempt to winback former subscribers, so retaining this data seems to have a valid business case

alexduf: That's a valid question, however:
- at the time we insert the link we don't have the subscription data yet (though it's not unsolvable)
- devices will regularly come to update that table, pretty much the same as we register devices for push notifications, so really it's 30 months since we've seen a user.

alexduf: Actually that raises an interesting corner case.

If a user purchase a subscription with Apple, then switch to using a different device (either different apple account on a different phone, or an Android Phone), then the user link won't be refreshed. So the TTL doesn't cover that case.

One possible work around might be to use the dynamo stream of the subscription table to know when a record is being deleted (30 months after the expiry date of the subscription) and delete the linked record on the user link table.

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-streams.html

alexduf: MSS-1081

PR #82, opened 2019-08-22
Prepare for prod

alexduf:
In this PR:

- Re-establish the environment checks to point PROD to PROD and CODE to Sandbox
- Load the apple "password" from the configuration, and delete it from all payloads.
- Re-use the SQS models between the link lambda and the rest of the codebase
- Refactor the link lambda to share more code between apple and google
- Generally fix the lambda as it wasn't doing what it was expected
- Start using async await in the link lambda. We should probably start using that pattern everywhere but I didn't feel like rewriting the codebase

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -104,6 +104,7 @@ Resources:
               Action:
                   - "dynamodb:GetItem"
                   - "dynamodb:BatchGetItem"
+                  - "dynamodb:BatchWriteItem"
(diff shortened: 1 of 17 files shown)


jacobwinch:
> * We should probably start using that pattern everywhere but I didn't feel like rewriting the codebase

I'm very glad you didn't decide to list the change:
* rewrite the codebase

In a PR called 'Prepare for prod' 😄

PR #84, opened 2019-08-27
Export subscription data to the datalake

alexduf:
This creates one lambda (bundled once, but used by two lambda functions).

That lambdas does a full scan of the table it receives as an argument, then streams the results to s3 as a big JSON.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -784,6 +784,38 @@ Resources:
       InsufficientDataActions:
         - Ref: AlarmTopic
 
+  ExportLambdasRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service:
+                - lambda.amazonaws.com
+            Action: sts:AssumeRole
+      Path: /
+      Policies:
+        - PolicyName: logs
+          PolicyDocument:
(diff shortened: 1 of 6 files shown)


alexduf:
Hmm, my lambda reports success but hasn't got the right to write on the bucket I have in CODE. The file isn't written.

Still the lambda reports success. So I need to understand what's up here

cloudformation.yaml
```diff
@@ -41,7 +41,12 @@ Parameters:
   AlarmTopic:
     Type: String
     Description: The ARN of the SNS topic to send all the cloudwatch alarms to
-
+  SubscriptionBucket:
```
NathanielBennett: Rename 'SubscriptionExportBucket'. Makes it clearer

cloudformation.yaml
```diff
@@ -41,7 +41,12 @@ Parameters:
...
+    Type: String
+    Description: The name of the export subscription bucket
+  UserSubscriptionBucket:
```
NathanielBennett: Ditto

NathanielBennett:
Just little things

alexduf:
Thanks, I've fixed the name of the cloudformation parameters.
I've also:

- Tested on CODE
- created the 4 associated s3 buckets in the ophan account, checked the data was exporting there

PR #89, opened 2019-09-02
Part two of exporting the events to the datalake

alexduf:
This PR aims to export the purchase events to the datalake RAW buckets.

- Drop the v1 table following the migration to the v2 of the subscription events table (v1 isn't used anymore at this point)
- Fix the date logic of the exports. The date in the s3 bucket should be yesterday. So on the 2nd of september at 00:01, what we really are exporting is the 1st of september
- Take an optional event as a parameter for the events export as we will want the capability to export previous days
- The actual logic of the export

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -47,6 +47,9 @@ Parameters:
   UserSubscriptionExportBucket:
     Type: String
     Description: The name of the export user subscription bucket
+  SubscriptionEventsExportBucket:
+    Type: String
+    Description: The name of the export events bucket
 
 Resources:
   MobilePurchasesLambdasRole:
@@ -816,9 +819,11 @@ Resources:
               Effect: Allow
               Action:
                 - "dynamodb:Scan"
+                - "dynamodb:Query"
               Resource:
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscriptions
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
+                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscription-events-v2/*
         - PolicyName: s3
           PolicyDocument:
             Statement:
@@ -829,6 +834,7 @@ Resources:
               Resource:
                 - !Sub arn:aws:s3:::${SubscriptionExportBucket}/*
                 - !Sub arn:aws:s3:::${UserSubscriptionExportBucket}/*
+                - !Sub arn:aws:s3:::${SubscriptionEventsExportBucket}/*
 
   ExportSubscriptionTableLambda:
     Type: AWS::Serverless::Function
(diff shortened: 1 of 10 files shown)


cloudformation.yaml
```diff
@@ -816,9 +819,11 @@ Resources:
               Effect: Allow
               Action:
                 - "dynamodb:Scan"
+                - "dynamodb:Query"
               Resource:
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscriptions
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
+                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-subscription-events-v2/*
```
alexduf: We need `/*` because we're not querying the table but the index

dynamo.cloudformation.yaml
```diff
@@ -96,14 +66,14 @@ Resources:
...
-        - IndexName: date-timestamp-index
+        - IndexName: date-timestamp-index-v2
           KeySchema:
             - AttributeName: date
               KeyType: HASH
             - AttributeName: timestamp
               KeyType: RANGE
           Projection:
-            ProjectionType: KEYS_ONLY
+            ProjectionType: ALL
```
alexduf: We need all the keys on that index

typescript/src/export/dynamoStream.ts
```diff
@@ -4,16 +4,24 @@ import {ScanIterator} from "@aws/dynamodb-data-mapper";
 export class DynamoStream<T> extends Readable {
 
     iterator: ScanIterator<T>;
+    transformItem: (t: T) => any;
 
-    constructor(iterator: ScanIterator<T>, opts?: ReadableOptions) {
+    constructor(iterator: ScanIterator<T>, transformItem?: (t: T) => any, opts?: ReadableOptions) {
```
alexduf: I've added this function to allow (one of) the export lambdas to sanitize the data before serializing it to file

typescript/src/models/subscriptionEvent.ts
```diff
@@ -40,4 +40,10 @@ export class SubscriptionEvent {
     get [DynamoDbTable]() {
         return App + "-" + Stage + "-subscription-events-v2";
     }
+}
+
+export class ReadSubscriptionEvent extends SubscriptionEvent {
+    constructor() {
```
alexduf: We need a zero parameter constructor for the mapper to automagically marshall the json

typescript/src/export/exportEvents.ts
```diff
@@ -0,0 +1,50 @@
...
+import {dynamoMapper, s3} from "../utils/aws";
+import zlib from 'zlib'
+import {Stage} from "../utils/appIdentity";
+import {DynamoStream} from "./dynamoStream";
+import {ReadSubscriptionEvent, SubscriptionEvent} from "../models/subscriptionEvent";
+import {plusDays} from "../utils/dates";
+
+function cleanupEvent(subEvent: SubscriptionEvent): any {
+    if (subEvent.applePayload) {
+        delete subEvent.applePayload.password; // just to be safe
```
jacobwinch: 🏅

PR #90, opened 2019-09-02
Various enhancements

alexduf:
- move the legacy table to being billed on demand rather than provisioned
- use inheritance to have a zero argument constructor on the userSubscription model
- simplify the webpack build

### DIFF
--- a/dynamo.cloudformation.yaml
+++ b/dynamo.cloudformation.yaml
@@ -2,12 +2,6 @@ AWSTemplateFormatVersion: 2010-09-09
 Description: 'Mobile Purchases DynamoDB'
 
 Parameters:
-  ReadCapacityUnits:
-    Description: DynamoDb Read Capacity Units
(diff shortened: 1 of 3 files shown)


dynamo.cloudformation.yaml
```diff
@@ -31,9 +25,7 @@ Resources:
       KeySchema:
         - AttributeName: userIdColonAppId
           KeyType: HASH
-      ProvisionedThroughput:
-        ReadCapacityUnits: !Ref ReadCapacityUnits
-        WriteCapacityUnits: !Ref WriteCapacityUnits
+      BillingMode: PAY_PER_REQUEST
```
jacobwinch: Do we expect to save $$$ here or just simplify things for ourselves?

alexduf: Both! although I don't expect the savings to be champaign worthy

PR #93, opened 2019-09-03
Link retention

alexduf:
The rows in the `UserSubscription` were deleted 30 months after the last interaction with a device.

This doesn't make sense, we need to delete these rows as soon as the subscription attached to it is being deleted.

This PR:

- Enable the dynamo event stream for the `Subscription` table
- creates a lambda that listen to that stream
- reacts only on TTL events from the `Subscription` table
- delete elements in the `UserSubscription` table when the corresponding row has been deleted from the `Subscription` table

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "es2018",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+    "target": "es2018",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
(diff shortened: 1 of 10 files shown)


tsconfig.json
```diff
@@ -11,10 +11,10 @@
     // "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
     "sourceMap": true,                        /* Generates corresponding '.map' file. */
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-     "outDir": "tsc-target",                      /* Redirect output structure to the directory. */
+    "outDir": "tsc-target",                   /* Redirect output structure to the directory. */
     // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     // "composite": true,                     /* Enable project compilation */
-    // "incremental": true,                   /* Enable incremental compilation */
+    "incremental": true,                      /* Enable incremental compilation */
```
NathanielBennett: What's the purpose of this change?

alexduf: Faster compilation time. Admittedly unrelated!

NathanielBennett: Fair play!

PR #97, opened 2019-09-11
Support more validation cases

alexduf:
There were two cases to support:

 - `latest_receipt_info` can either be an array or a value
 - `expires_date` sometimes contains a formatted date "2019-09-11 00:00 Timezone" or sometimes a string representation of a timestamp "1567081703000"

And I thought our own APIs were dirty but Apple raises the bar quite high.

I've added unit tests because I suspect that's not the last surprise so we may need to add more unit tests as we go.

### DIFF
--- a/typescript/src/services/appleValidateReceipts.ts
+++ b/typescript/src/services/appleValidateReceipts.ts
@@ -13,13 +13,23 @@ export interface AppleValidatedReceiptInfo {
 }
 
 // there are more fields, I cherry picked what was relevant
-// https://developer.apple.com/documentation/appstorereceipts/responsebody
-export interface AppleValidationResponse {
+// https://developer.apple.com/library/archive/releasenotes/General/ValidateAppStoreReceipt/Chapters/ValidateRemotely.html
+interface AppleValidationServerResponse {
     auto_renew_status: 0 | 1,
     "is-retryable": boolean,
     latest_receipt: string,
-    latest_receipt_info: AppleValidatedReceiptInfo,
+    // yes you've read the type well. It can both be an array or a value, good luck parsing that
+    latest_receipt_info: AppleValidatedReceiptInfo | AppleValidatedReceiptInfo[],
+    latest_expired_receipt_info: AppleValidatedReceiptInfo,
     status: number
+}
+
+// this is a sanitised and more sensible version of what the response should be
+export interface AppleValidationResponse {
+    autoRenewStatus: boolean,
+    isRetryable: boolean,
+    latestReceipt: string,
+    latestReceiptInfo: AppleValidatedReceiptInfo
 }
 
 const receiptEndpoint = (Stage === "PROD") ? "https://buy.itunes.apple.com/verifyReceipt" : "https://sandbox.itunes.apple.com/verifyReceipt";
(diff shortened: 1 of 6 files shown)


typescript/src/services/appleValidateReceipts.ts
```diff
@@ -3,25 +3,46 @@ import {getConfigValue} from "../utils/ssmConfig";
...
-export interface AppleValidationResponse {
+// https://developer.apple.com/library/archive/releasenotes/General/ValidateAppStoreReceipt/Chapters/ValidateRemotely.html
+export interface AppleValidationServerResponse {
     auto_renew_status: 0 | 1,
-    "is-retryable": boolean,
-    latest_receipt: string,
-    latest_receipt_info: AppleValidatedReceiptInfo,
+    "is-retryable"?: boolean,
+    latest_receipt?: string,
+    // yes you've read the type well. It can both be an array or a value, good luck parsing that
```
jacobwinch: 😿

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -42,25 +63,86 @@ function callValidateReceipt(receipt: string): Promise<Response> {
...
 }
 
-function checkResponseStatus(response: AppleValidationResponse): AppleValidationResponse {
+function checkResponseStatus(response: AppleValidationServerResponse): AppleValidationServerResponse {
     if ((response.status >= 21100 && response.status <= 21199) || response["is-retryable"]) {
         console.error(`Server error received from Apple, got status ${response.status}, will retry`);
         throw new ProcessingError(`Server error, status ${response.status}`, true);
     }
+    if (response.status == 21007 || response.status == 21008) {
+        console.error(`Got status ${response.status} and we're in ${Stage}, so we are processing a receipt from the wrong environment`);
```
jacobwinch: This is very helpful logging 👍

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -3,25 +3,46 @@ import {getConfigValue} from "../utils/ssmConfig";
...
+    cancellationDate: Option<Date>,
+    expiresDate: Date,
+    originalPurchaseDate: Date,
+    originalTransactionId: string
+    productId: string,
+}
+
+// this is a sanitised and more sensible version of what the response should be
+export interface AppleValidationResponse {
+    autoRenewStatus: boolean,
```
jacobwinch: Minor: could drop the word `status` here

alexduf: It's true but I wanted to keep the field named the same way as Apple

typescript/tests/services/appleValidateReceipts.test.ts
```diff
@@ -0,0 +1,132 @@
...
+                originalTransactionId: "1234",
+                productId: "uk.co.guardian.gla.1month.2018May.withFreeTrial",
+            },
+        };
+
+        expect(toSensiblePayloadFormat(appleResponse, "cmVjZWlwdA==")).toStrictEqual(expected);
+    });
+
+
+    test("Should transform a dirty apple payload with an expired receipt info into a sane one", () => {
```
jacobwinch: We might want to use a slightly different description here (for clarity)?

alexduf: Good shout

typescript/src/services/appleValidateReceipts.ts
jacobwinch: Ah yes, I see that now - fair enough 👍

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -42,25 +63,86 @@ function callValidateReceipt(receipt: string): Promise<Response> {
...
+    }
+
+    let receiptInfo: AppleValidatedReceiptServerInfo;
+    if (response.latest_receipt_info) {
+        if (Array.isArray(response.latest_receipt_info)) {
+            const latestReceipt = response.latest_receipt_info as AppleValidatedReceiptServerInfo[];
+            if (latestReceipt.length == 1) {
+                receiptInfo = latestReceipt[0];
+            } else if (latestReceipt.length > 1) {
+                receiptInfo = latestReceipt[0];
```
jacobwinch: I think we should be able to replace this block with something like:

```receiptInfo = latestReceipt.sort(receiptInfo => expiryDate(receiptInfo))[0]```

Comment (alexduf) [reply typescript/src/services/appleValidateReceipts.ts]: That's a really good shout

File: typescript/src/services/appleValidateReceipts.ts [hunk 6]
```diff
@@ -42,25 +63,86 @@ function callValidateReceipt(receipt: string): Promise<Response> {
...
+
+    function expiryDate(receiptServerInfo: AppleValidatedReceiptServerInfo): number {
+        if (receiptServerInfo.expires_date_ms) {
+            return Number.parseInt(receiptServerInfo.expires_date_ms);
+        } else {
+            return Number.parseInt(receiptServerInfo.expires_date);
+        }
+    }
+
+    let receiptInfo: AppleValidatedReceiptServerInfo;
```
jacobwinch: I think it might be cleaner to pull all of this into a function `getReceiptInfo`, which could probably also allow us to avoid using `let`?

PR #101, opened 2019-09-25
API key & Payload

alexduf:
- Access from an API key, providing user ID in the URL
 - Change the payload returned by the user endpoint

This is part of the work to open this API to the members data api

### DIFF
--- a/typescript/src/link/link.ts
+++ b/typescript/src/link/link.ts
@@ -3,7 +3,7 @@ import {HTTPResponses} from '../models/apiGatewayHttp';
 import {UserSubscription} from "../models/userSubscription";
 import {ReadSubscription} from "../models/subscription";
 import {dynamoMapper, sqs} from "../utils/aws";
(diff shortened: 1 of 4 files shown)


typescript/src/link/link.ts
```diff
@@ -62,15 +62,19 @@ export async function parseAndStoreLink<A, B>(
...
-            console.log(`Put ${insertCount} links in the DB, and sent ${sqsCount} subscription refs to SQS`);
+            if (userId) {
+                const insertCount = await persistUserSubscriptionLinks(toUserSubscription(userId, payload));
+                const sqsCount = await enqueueUnstoredPurchaseToken(toSqsPayload(payload));
+                console.log(`Put ${insertCount} links in the DB, and sent ${sqsCount} subscription refs to SQS`);
 
-            return HTTPResponses.OK;
+                return HTTPResponses.OK;
+            } else {
+                return HTTPResponses.UNAUTHORISED;
```
NathanielBennett: Yeah this is better. :)

PR #104, opened 2019-10-15
Move exports into their own cloudformation template

alexduf:
I'll have to manually create the cloudformation stacks.

I've done no modification to the actual template, just moved code and duplicated parameters when necessary.


exports-cloudformation.yaml
```diff
@@ -0,0 +1,166 @@
...
+  Stage:
+    Description: Stage name
+    Type: String
+    AllowedValues:
+      - CODE
+      - PROD
+  DeployBucket:
+    Description: Bucket where RiffRaff uploads artifacts on deploy
+    Type: String
+  AlarmTopic:
```
jacobwinch: I don't think this parameter is used in the new exports template?

PR #112, opened 2019-11-15
Support checking subscription status for multiple Android apps

jacobwinch:
The team developing the puzzles app intend to start using this API to check Play subscription status.

This PR makes it possible for clients to check the subscription status of any app in our account by passing the relevant package name as a header. We default to `com.guardian` if this header is not present, which means that existing clients (i.e. current versions of the Guardian Live App) should continue to work without providing this header.

### DIFF
--- a/typescript/src/subscription-status/googleSubStatus.ts
+++ b/typescript/src/subscription-status/googleSubStatus.ts
@@ -41,7 +41,10 @@ export async function handler(request: APIGatewayProxyEvent): Promise<APIGateway
     if (request.pathParameters && request.headers && getPurchaseToken(request.headers)) {
         const restClient = new restm.RestClient('guardian-mobile-purchases');
         const purchaseToken = getPurchaseToken(request.headers);
-        const url = buildGoogleUrl(request.pathParameters.subscriptionId, purchaseToken, googlePackageName(request.headers));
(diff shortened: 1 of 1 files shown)


typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -25,15 +25,26 @@ function getPurchaseToken(headers: HttpRequestHeaders): string {
...
     const stage = process.env.Stage;
-    const googlePackagename =  "com.guardian"
 
     if (request.pathParameters && request.headers && getPurchaseToken(request.headers)) {
         const restClient = new restm.RestClient('guardian-mobile-purchases');
-        const purchaseToken = getPurchaseToken(request.headers)
-        const url = buildGoogleUrl(request.pathParameters.subscriptionId, purchaseToken, googlePackagename);
+        const purchaseToken = getPurchaseToken(request.headers);
+        const packageName = googlePackageName(request.headers);
+        const subscriptionId = request.pathParameters.subscriptionId;
```
jacobwinch: In this context `subscriptionId` is something like `com.guardian.subscription.monthly.10`, which refers to a specific Play subscription configured in our account. It does not identify a user so it's safe to log IMO.

PR #113, opened 2019-11-18
Grace period, trial period and auto renewal

alexduf:
This PR adds the following features: 
 - Expose the end date including the grace period
 - Expose the trial period flag as given by apple
 - Expose the auto renewal status flag as given by apple
 - Move the auto renewal status flag at the receipt level, and change the logic to match it with Apple's PendingRenewalInfo object based on the original transaction ID

### DIFF
--- a/typescript/tests/services/appleValidateReceipts.test.ts
+++ b/typescript/tests/services/appleValidateReceipts.test.ts
@@ -9,6 +9,7 @@ describe("The apple validation service", () => {
         const appleResponse: AppleValidationServerResponse = {
             auto_renew_status: 0,
             latest_expired_receipt_info: {
+                is_trial_period: "false",
                 original_transaction_id: "1234",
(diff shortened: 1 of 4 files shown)


typescript/tests/services/appleValidateReceipts.test.ts
```diff
@@ -190,4 +204,77 @@ describe("The apple validation service", () => {
...
+                    is_in_billing_retry_period: '0',
+                    product_id: 'uk.co.guardian.gla.6months.2018May.withFreeTrial',
+                    auto_renew_status: '1'
+                }
+            ]
+        };
+
+        const expected: AppleValidationResponse[] = [{
+            isRetryable: false,
+            latestReceipt: "cmVjZWlwdA==",
```
JamieB-gu: Ok to expose this publicly?

alexduf: Yes, this is the string `receipt` that has been base64 encoded

PR #114, opened 2019-11-19
Typescript 3.7

alexduf:
I tried to catch all the obvious uses of `variable || "defaultValue"` and of quasi false value in conditions, but I'm sure it's not exhaustive

### DIFF
--- a/typescript/src/link/deleteLink.ts
+++ b/typescript/src/link/deleteLink.ts
@@ -24,17 +24,14 @@ async function deleteUserSubscription(subscriptionId: string): Promise<number> {
 export async function handler(event: DynamoDBStreamEvent): Promise<any> {
     const ttlEvents = event.Records.filter(dynamoEvent => {
         return dynamoEvent.eventName == "REMOVE" &&
-            dynamoEvent.userIdentity &&
-            dynamoEvent.userIdentity.type == "Service" &&
-            dynamoEvent.userIdentity.principalId == "dynamodb.amazonaws.com" &&
-            dynamoEvent.dynamodb &&
-            dynamoEvent.dynamodb.Keys &&
-            dynamoEvent.dynamodb.Keys.subscriptionId
+            dynamoEvent.userIdentity?.type == "Service" &&
+            dynamoEvent.userIdentity?.principalId == "dynamodb.amazonaws.com" &&
+            dynamoEvent.dynamodb?.Keys?.subscriptionId
     });
 
     const subscriptionsPromises = ttlEvents
         // @ts-ignore
-        .map(event => event.dynamodb.Keys.subscriptionId.S || "")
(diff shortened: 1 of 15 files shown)


typescript/src/link/deleteLink.ts
```diff
@@ -24,17 +24,14 @@ async function deleteUserSubscription(subscriptionId: string): Promise<number> {
...
         return dynamoEvent.eventName == "REMOVE" &&
-            dynamoEvent.userIdentity &&
-            dynamoEvent.userIdentity.type == "Service" &&
-            dynamoEvent.userIdentity.principalId == "dynamodb.amazonaws.com" &&
-            dynamoEvent.dynamodb &&
-            dynamoEvent.dynamodb.Keys &&
-            dynamoEvent.dynamodb.Keys.subscriptionId
+            dynamoEvent.userIdentity?.type == "Service" &&
+            dynamoEvent.userIdentity?.principalId == "dynamodb.amazonaws.com" &&
+            dynamoEvent.dynamodb?.Keys?.subscriptionId
```
JamieB-gu: Worth converting `==` to `===`?

typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -22,11 +22,11 @@ interface SubscriptionStatusResponse {
 }
 
 function getPurchaseToken(headers: HttpRequestHeaders): string {
-    return headers["Play-Purchase-Token"] || headers["play-purchase-token"]
+    return headers["Play-Purchase-Token"] ?? headers["play-purchase-token"]
```
JamieB-gu: What happens if `headers["play-purchase-token"]` is missing?

typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -22,11 +22,11 @@ interface SubscriptionStatusResponse {
...
 }
 
 function googlePackageName(headers: HttpRequestHeaders): string {
-    const packageNameFromHeaders = headers["Package-Name"] || headers["package-name"];
+    const packageNameFromHeaders = headers["Package-Name"] ?? headers["package-name"];
```
JamieB-gu: As above ^

typescript/src/subscription-status/googleSubStatus.ts
alexduf: The function returns null, and the server returns an HTTP 400 invalid request (see line 77)

typescript/src/subscription-status/googleSubStatus.ts
alexduf: line 33, we test for nullity of the value

typescript/src/subscription-status/googleSubStatus.ts
JamieB-gu: If the function returns `null`, should that be reflected in the return type?

PR #117, opened 2019-11-26
Trace if the server and the client disagree on the validity of the receipt

alexduf:
As per our chat with Sophie


SLambrakis:
I'll leave the actual approval to someone else, but the logic looks sound to me. Thanks for this!

PR #121, opened 2019-12-20
MSS-1345: Fix jackson-databind vulnerability

frankie297:
Force jackson-databind version to fix some vulnerability issues:
https://app.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-174736

NB: remove after upgrading play app to 2.7 (if we should ever do this)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,7 @@ import scala.collection.immutable
 val testAndCompileDependencies: String = "test->test;compile->compile"
 val awsVersion: String = "1.11.375"
 val simpleConfigurationVersion: String = "1.4.3"
+
 // Force a version of jackson-databind that addresses this vulnerability:
 // https://app.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-471943
 // introduced via com.typesafe.play:play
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -7,13 +7,17 @@ import scala.collection.immutable
 val testAndCompileDependencies: String = "test->test;compile->compile"
 val awsVersion: String = "1.11.375"
 val simpleConfigurationVersion: String = "1.4.3"
+val jacksonData: String = "2.9.10.1"
+val jacksonDatatype: String = "2.9.8"
```
jacobwinch: Is `jacksonDatatype` used?

build.sbt
```diff
@@ -7,13 +7,17 @@ import scala.collection.immutable
 val testAndCompileDependencies: String = "test->test;compile->compile"
 val awsVersion: String = "1.11.375"
 val simpleConfigurationVersion: String = "1.4.3"
+val jacksonData: String = "2.9.10.1"
```
jacobwinch: It might be worth adding a comment here to explain why we're pinning the `jackson-databind` version?

build.sbt
frankie297: good point, i don't think it is needed. Thanks for spotting (i might have just carried this over from another repo as I have done the same change multiple times) 👍

frankie297:
@jacobwinch Just so you are aware I have actually upgraded to a newest version due to a recent vulnerability in `2.9.10.1`

build.sbt
frankie297: 👍

PR #124, opened 2020-01-22
Fallback on the sandbox environment

alexduf:
This is to allow beta build to validate their receipts without having to change endpoint.

The apple receipt validation service will retry to validate a receipt if it receives the code 21007, which indicates that a sandbox receipt has been validated in prod.

We only want to retry when validating a receipt as a one off. The process that stores the state of subscription shouldn't try to revalidate receipts.

### DIFF
--- a/typescript/src/services/appleValidateReceipts.ts
+++ b/typescript/src/services/appleValidateReceipts.ts
@@ -26,6 +26,10 @@ export interface AppleValidatedReceiptServerInfo {
     is_trial_period: string
 }
 
+export interface ValidationOptions {
+    sandboxRetry: boolean
+}
+
 // there are more fields, I cherry picked what was relevant
 // https://developer.apple.com/library/archive/releasenotes/General/ValidateAppStoreReceipt/Chapters/ValidateRemotely.html
 export interface AppleValidationServerResponse {
@@ -181,9 +185,9 @@ export function toSensiblePayloadFormat(response: AppleValidationServerResponse,
     })
 }
 
-async function retryInSandboxIfNecessary(parsedResponse: AppleValidationServerResponse, receipt: string): Promise<AppleValidationServerResponse> {
-    if (parsedResponse.status === 21007) {
-        console.log("Got return code 21007, retrying in Sandbox");
+async function retryInSandboxIfNecessary(parsedResponse: AppleValidationServerResponse, receipt: string, options: ValidationOptions): Promise<AppleValidationServerResponse> {
+    if (parsedResponse.status === 21007 && options.sandboxRetry) {
+        console.log("Got status code 21007, retrying in Sandbox");
         return callValidateReceipt(receipt, true)
             .then(response => response.json())
             .then(body => body as AppleValidationServerResponse);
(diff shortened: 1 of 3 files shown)


typescript/src/services/appleValidateReceipts.ts
```diff
@@ -57,17 +61,21 @@ export interface AppleValidationResponse {
...
             const body = JSON.stringify({
                 "receipt-data": receipt,
                 "password": password,
                 "exclude-old-transactions": true
             });
-            return fetch(receiptEndpoint, { method: 'POST', body: body});
+            return fetch(endpoint, { method: 'POST', body: body});
         }).then(response => {
             if (!response.ok) {
                 console.error(`Impossible to validate the receipt, got ${response.status} ${response.statusText} from receiptEndpoint for ${receipt}`);
```
jacobwinch: Should this be `${endpoint}` now?

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -57,17 +61,21 @@ export interface AppleValidationResponse {
     originalResponse: any
 }
 
-const receiptEndpoint = (Stage === "PROD") ? "https://buy.itunes.apple.com/verifyReceipt" : "https://sandbox.itunes.apple.com/verifyReceipt";
 
-function callValidateReceipt(receipt: string): Promise<Response> {
+const sandboxReceiptEndpoint = "https://sandbox.itunes.apple.com/verifyReceipt";
+const prodReceiptEndpoint = "https://buy.itunes.apple.com/verifyReceipt";
+const receiptEndpoint = (Stage === "PROD") ? sandboxReceiptEndpoint : prodReceiptEndpoint;
```
jacobwinch: I think this should be `const receiptEndpoint = (Stage === "PROD") ? prodReceiptEndpoint : sandboxReceiptEndpoint;`?

jacobwinch: (I think you've accidentally reversed the order when pulling these urls into consts)

typescript/src/subscription-status/appleSubStatus.ts
```diff
@@ -12,7 +12,7 @@ interface AppleSubscription {
 }
 
 interface AppleLinkPayload {
-    platform: Platform.DailyEdition | Platform.Ios
+    platform: Platform.DailyEdition | Platform.Ios | Platform.IosPuzzles
```
jacobwinch: One for another PR, but platform isn't really an appropriate name for this type anymore!

typescript/src/services/appleValidateReceipts.ts
alexduf: You're right 😨

typescript/src/subscription-status/appleSubStatus.ts
alexduf: I guess you're right! unfortunately the column is named platform so I think I'd rather live with it :/

typescript/src/update-subs/apple.ts
```diff
@@ -36,7 +36,7 @@ function toAppleSubscription(response: AppleValidationResponse): Subscription {
 function sqsRecordToAppleSubscription(record: SQSRecord): Promise<Subscription[]> {
     const subRef = JSON.parse(record.body) as AppleSubscriptionReference;
 
-    return validateReceipt(subRef.receipt)
+    return validateReceipt(subRef.receipt, {sandboxRetry: false})
```
jacobwinch: Does Apple send us sandbox events via server to server notifications? If so, presumably this is important as it prevents us from polluting our prod tables with sandbox data? If I've understood that correctly, it might be worth explicitly calling this out here?

typescript/src/subscription-status/appleSubStatus.ts
jacobwinch: Oh dear, yes I think we can blame it on `legacy reasons` and leave it as is then!

typescript/src/update-subs/apple.ts
alexduf: >Does Apple send us sandbox events via server to server notifications? 
No, only production events

>presumably this is important as it prevents us from polluting our prod tables with sandbox data
you're right I've added a comment to explain why

PR #125, opened 2020-01-27
Add a global secondary index to subscription table

frankie297:
Creates a new global secondary index in the subscription table, 

@alexduf the only thing I am unsure is whether to include `ALL` the attributes or just to `INCLUDE` the `latest_receipt` and the `autoRenewStatus` attributes.

### DIFF
--- a/dynamo.cloudformation.yaml
+++ b/dynamo.cloudformation.yaml
@@ -102,6 +102,15 @@ Resources:
         AttributeName: ttl
       StreamSpecification:
         StreamViewType: KEYS_ONLY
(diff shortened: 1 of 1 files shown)


alexduf:
Ideally it would only include a set of attributes in the projection rather than all the attributes

frankie297:
> Great! have you checked it on CODE?

Showing in CODE ->
Screenshot 2020-01-30 at 18 00 42

N.B: I had to add an `endTimestamp` attribute name in the attribute definitions for the template to pass

PR #126, opened 2020-02-03
More capacity for subscription export

alexduf:
The subscription export job has failed with a timeout a few times recently. Looking at the trends it seems like the duration is proportional to the number of rows.  Since we've had more and more rows the export has now reached the maximum 15 minutes a lambda can run for.

In the short term, I'm quadrupling the capacity of the lambda it as it seems to have an impact on how quickly the table can be exported (I tested with doubling it)

In the long term we may need to re-think this export, as a lambda + s3 might not be enough to fit within 15 minutes.

### DIFF
--- a/exports-cloudformation.yaml
+++ b/exports-cloudformation.yaml
@@ -110,7 +110,7 @@ Resources:
           ExportBucket: !Ref SubscriptionExportBucket
           ClassName: ReadSubscription
       Description: Export subscription table to the datalake
-      MemorySize: 512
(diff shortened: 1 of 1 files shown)


alexduf:
As a note for later, 512MB of RAM (and it corresponding CPU) would do the export in about 900s
1024MB does it in 552s, and 2048 does it in 360s

It's not quite a linear correlation as quadrupling the capacity divides the processing time by 3, but good enough

PR #128, opened 2020-02-05
Create filter for subscription table

frankie297:
Select all the receipts in the ios-endTimestamp-revalidation-index index, filtered on 
- auto-renewing subscription (true)
- that contain a receipt
- where endTimestamp parameter is before endTimestampFilter

Related to:
https://github.com/guardian/mobile-purchases/pull/127
https://github.com/guardian/mobile-purchases/pull/130

This has been tested locally,
Reminder: the cloudformation template needs to be updated in the console for CODE and PROD

This was a joint PR with @jacobwinch

### DIFF
--- a/typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
+++ b/typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
@@ -1,18 +1,18 @@
 import {plusHours} from "../utils/dates";
 import {dynamoMapper} from "../utils/aws";
 import {endTimeStampFilterSubscription} from "../models/endTimestampFilter";
-import {equals} from '@aws/dynamodb-expressions';
-import {ReadUserSubscription} from "../models/userSubscription";
-import {TableName} from "aws-sdk/clients/dynamodb";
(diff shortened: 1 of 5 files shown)


typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
```diff
@@ -1,18 +1,56 @@
...
+import {dynamoMapper} from "../utils/aws";
+import {endTimeStampFilterSubscription} from "../models/endTimestampFilter";
+import {
+ AndExpression,
+ attributeExists,
+ equals, lessThan,
+} from '@aws/dynamodb-expressions';
 
-function endTimestampForQuery(event: ScheduleEvent): Date {
+function endTimestampForQuery(event: ScheduleEvent): string {
```
alexduf: It's a bit philosophical but here it might have been nice to keep the return type as a `Date`, and call `.toISOString()` on that date when the function is called (line 24).

Nothing blocking, but it's just a little bit nicer as it guarantees that future modification of this function will always return a date, so no-one returns `'a'` by mistake

PR #131, opened 2020-02-13
Add receipts that need to be revalidated to SQS queue

frankie297:
Add all the receipts that have been filtered and need to be revalidated to the `apple-subscriptions-fetch queue`

@alexduf I wasn't sure which action is more appropriate?
`              - "sqs:SendMessage"`
               `- "sqs:SendMessageBatch"`

This has been tested on CODE, messages were received on the queue and no messages were in the DLQ

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -978,6 +978,11 @@ Resources:
       Description: Finds recently expired subscriptions
       MemorySize: 512
       Timeout: 60
+      Events:
+        Schedule:
+          Type: Schedule
+          Properties:
+            Schedule: cron(0 0 0/6 ? * * *)
       Tags:
         Stage: !Ref Stage
         Stack: !Ref Stack
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -887,6 +887,15 @@ Resources:
               Resource:
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions/*
+        - PolicyName: sqs
+          PolicyDocument:
+            Statement:
+              Effect: Allow
+              Action:
+                - "sqs:SendMessage"
+                - "sqs:SendMessageBatch"
+              Resource:
+                - !Sub arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:${App}-${Stage}-apple-subscriptions-to-fetch
 
   AppleRevalidateSubscriptionRole:
     Type: AWS::IAM::Role
@@ -966,6 +975,7 @@ Resources:
           App: !Sub ${App}
           Stack: !Sub ${Stack}
           Stage: !Sub ${Stage}
+          SqsUrl: !Ref AppleSubscriptionsQueue
       Description: Finds recently expired subscriptions
       MemorySize: 512
       Timeout: 60
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -893,7 +893,6 @@ Resources:
               Effect: Allow
               Action:
                 - "sqs:SendMessage"
-                - "sqs:SendMessageBatch"
               Resource:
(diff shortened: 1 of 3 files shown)


typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
```diff
@@ -50,7 +50,10 @@ export async function handler(event: ScheduleEvent) {
  });
 
  for await (const subscription of queryScan) {
+  const queueUrl = process.env.SqsUrl
+  if (queueUrl === undefined) throw new Error("No QueueUrl env parameter provided");
```
alexduf: You seem to have named your environment variable `SqsUrl`, yet this message refers to it as `QueueUrl`

cloudformation.yaml
```diff
@@ -887,6 +887,15 @@ Resources:
               Resource:
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
                 - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions/*
+        - PolicyName: sqs
+          PolicyDocument:
+            Statement:
+              Effect: Allow
+              Action:
+                - "sqs:SendMessage"
+                - "sqs:SendMessageBatch"
```
alexduf: To answer your question you can check which function is being called [here](https://github.com/guardian/mobile-purchases/blob/d74dc3c5d8ab36797967f72ba4cc8c9270a10674/typescript/src/utils/aws.ts#L42)

So only one of the two is necessary :)

typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
```diff
@@ -50,7 +50,10 @@ export async function handler(event: ScheduleEvent) {
...
+
+  await sendToSqs(queueUrl, queryScan)
```
alexduf: Here you have to wonder what you want to send to the SQS queue. Is `queryScan` really what you want to send?

If you're not sure of what to send, maybe you can look at it the other way around by checking what the lambda [reading the queue expects](https://github.com/guardian/mobile-purchases/blob/8de5a8a8df4a7b8b01812f972e99b92525ae5ed1/typescript/src/update-subs/apple.ts#L37)?

frankie297: Screenshot 2020-02-14 at 12 45 43
This is the output now, I wasn't sure how to deal with the `else`, it's currently a `console.log` should it be an `error`?

typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
```diff
@@ -1,56 +1,67 @@
...
-  {
-   indexName: 'ios-endTimestamp-revalidation-index',
-   filter: filter
- });
+    const queryScan = dynamoMapper.scan(
+        EndTimeStampFilterSubscription,
+        {
+            indexName: 'ios-endTimestamp-revalidation-index',
+            filter: filter,
+            limit: 1
```
alexduf: You probably want to remove that one :)

typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
```diff
@@ -1,56 +1,67 @@
...
+import {AppleSubscriptionReference} from "../models/subscriptionReference";
 
 function endTimestampForQuery(event: ScheduleEvent): Date {
- const defaultDate = plusHours(new Date(), 3);
- if(event.endTimestampFilter) {
-  return new Date(Date.parse(event.endTimestampFilter));
- } else {
-  return defaultDate;
- }
+    const defaultDate = plusHours(new Date(), 3);
```
alexduf: Maybe we should put 13 hours here, since the lambda is going to run every six hours, we're sure to try to validate the receipt at least twice before it's expired?

typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
frankie297: 👍

typescript/src/revalidate-receipts/appleRevalidateReceipts.ts
frankie297: 👍

PR #133, opened 2020-02-14
Enhance timeout handling

alexduf:
- Lambdas called by the API had a timeout of 45s, when the API Gateway has a default timeout of 30s. So I've set all the lambdas called by the API Gateway to timeout at 29s. This will allow us to know when a lambda timeouts a bit better has there was the possibility of seeing everything going well in the logs, yet the api gateway returning HTTP 504 (Timeout).

- Set a timeout on the request that goes to Apple to validate receipts such that we know if Apple is struggling to validate our receipts.

### DIFF
--- a/typescript/src/services/appleValidateReceipts.ts
+++ b/typescript/src/services/appleValidateReceipts.ts
@@ -75,7 +75,12 @@ function callValidateReceipt(receipt: string, forceSandbox: boolean = false): Pr
                 "password": password,
                 "exclude-old-transactions": true
             });
-            return fetch(endpoint, { method: 'POST', body: body});
+            return Promise.race([
+                fetch(endpoint, { method: 'POST', body: body}),
+                new Promise((_, reject) =>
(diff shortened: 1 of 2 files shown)


typescript/src/services/appleValidateReceipts.ts
```diff
@@ -75,7 +75,12 @@ function callValidateReceipt(receipt: string, forceSandbox: boolean = false): Pr
                 "password": password,
                 "exclude-old-transactions": true
             });
-            return fetch(endpoint, { method: 'POST', body: body});
+            return Promise.race([
+                fetch(endpoint, { method: 'POST', body: body}),
+                new Promise((_, reject) =>
+                    setTimeout(() => reject(new Error("Timeout while validating receipt with Apple")), 12000)
```
davidfurey: 12 second timeout?

alexduf: As generous as can be within a 29s window

PR #140, opened 2020-02-20
Fix where the bundle ID is read from.

alexduf:
So...

`bundle_id` which is [documented](https://developer.apple.com/library/archive/releasenotes/General/ValidateAppStoreReceipt/Chapters/ReceiptFields.html#//apple_ref/doc/uid/TP40010573-CH106-SW1) as the first field here, isn't always returned by Apple's service. Note how it's not mentioned that the field is optional.

There's also this [documentation](https://developer.apple.com/documentation/appstorereceipts/responsebody/receipt), which is more recent, does mention the `bundle_id`, but doesn't mention that it's optional.

What I'm seeing in prod (but not in sandbox that would be too easy) is this field isn't present, but there's another field! called `bid` which behaves the same way. `bid` is undocumented.

Apple says:

> Keys not documented below are reserved for use by Apple and must be ignored by your app.

Yes, I'd very much like that too, but you leave me no choice Apple...

### DIFF
--- a/typescript/src/services/appleValidateReceipts.ts
+++ b/typescript/src/services/appleValidateReceipts.ts
@@ -181,6 +181,9 @@ export function toSensiblePayloadFormat(response: AppleValidationServerResponse,
         // which is valued but undocumented. Oh and also sometimes it's on the latest_receipt object, but sometimes
         // it's on the receipt object. Hopefully this covers all the corner cases? who knows!
         const bundleId = receiptInfo.bundle_id ?? receiptInfo.bid ?? response.receipt?.bundle_id ?? response.receipt?.bid;
+        if (!bundleId) {
+            console.warn(`Unable to identify the bundle id for the original transaction id ${receiptInfo.original_transaction_id}`)
+        }
 
         return {
             isRetryable: response["is-retryable"] === true,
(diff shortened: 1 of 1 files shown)


SLambrakis:
😬 This definitely feels like something we should open a radar for. Would you like to draft an explanation and we can sit together next week to submit it?

alexduf:
Yes that sounds like a lovely idea Sophie, thanks 👍

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -176,11 +177,16 @@ export function toSensiblePayloadFormat(response: AppleValidationServerResponse,
         const pendingRenewalInfo: PendingRenewalInfo = pendingRenewalInfoById[receiptInfo.original_transaction_id];
         const autoRenewStatus = pendingRenewalInfo ? pendingRenewalInfo.auto_renew_status === "1" : response.auto_renew_status === 1;
 
+        // bundle_id is the documented field, however it's sometimes not valued, but instead there's the field bid
+        // which is valued but undocumented. Oh and also sometimes it's on the latest_receipt object, but sometimes
+        // it's on the receipt object. Hopefully this covers all the corner cases? who knows!
+        const bundleId = receiptInfo.bundle_id ?? receiptInfo.bid ?? response.receipt?.bundle_id ?? response.receipt?.bid;
```
webb04: Should we handle the case where it doesn't have anything? by the look if it that might happen 😅

alexduf: If we can't find it we'll leave the field unvalued as it is today. I guess I could add something to the logs?

alexduf: If that's of any help, the type of bundleId is `string | undefined`

PR #141, opened 2020-02-20
Use the rest client everywhere

alexduf:
`node-fetch` was only used in one spot, so we can drop one library and use the same HTTP client everywhere.

I also create one instance of the client in memory so that some lambda don't continuously re-create the HTTP client.

### DIFF
--- a/typescript/src/services/appleValidateReceipts.ts
+++ b/typescript/src/services/appleValidateReceipts.ts
@@ -1,10 +1,10 @@
-import {Response} from "node-fetch";
 import {getConfigValue} from "../utils/ssmConfig";
 import {ProcessingError} from "../models/processingError";
 import {Stage} from "../utils/appIdentity";
-import fetch from 'node-fetch';
 import {msToDate, optionalMsToDate} from "../utils/dates";
 import {Option} from "../utils/option";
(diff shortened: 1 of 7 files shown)


typescript/src/services/appleValidateReceipts.ts
```diff
@@ -78,15 +77,12 @@ function callValidateReceipt(receipt: string, forceSandbox: boolean = false): Pr
                 "password": password,
                 "exclude-old-transactions": true
             });
-            return Promise.race([
-                fetch(endpoint, { method: 'POST', body: body}),
-                new Promise((_, reject) =>
-                    setTimeout(() => reject(new Error("Timeout while validating receipt with Apple")), 12000)
-                ) as Promise<Response>
-            ]);
+            return restClient.client.post(endpoint, body)
```
alexduf: Here I'm using the underlying HTTP client of `restClient` as it's not really a restful call

alexduf:
Thank you!

typescript/src/services/appleValidateReceipts.ts
JamieB-gu: What happens to the 12 second timeout here? Is it replaced by the 10 second timeout you apply to the global rest client?

PR #142, opened 2020-02-21
Add the trial period flag (part 1)

alexduf:
This PR focuses solely on the refactoring necessary to add the trial period flag to the subscription table.

Each commit should be self explanatory:

- Remove dead code
- Move a model definition closer to where it's used (only used in one place)
- Refactor the subscription and update function to share code


typescript/src/services/google-play.ts
```diff
@@ -34,3 +35,18 @@ export function buildGoogleUrl(subscriptionId: String, purchaseToken: String, pa
...
 }
 
+export interface GoogleResponseBody {
+    startTimeMillis: string,
+    expiryTimeMillis: string,
+    userCancellationTimeMillis: string,
+    autoRenewing: boolean
+}
+
+export async function fetchGoogleSubscription(subscriptionId: string, purchaseToken: String, packageName: String): Promise<GoogleResponseBody | null> {
```
JamieB-gu:
```suggestion
export async function fetchGoogleSubscription(subscriptionId: string, purchaseToken: string, packageName: string): Promise<GoogleResponseBody | null> {
```

typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -3,77 +3,63 @@ import {
...
-                        console.log(`Serving an Internal Server Error due to: ${error}`);
-                        return HTTPResponses.INTERNAL_ERROR
-                    }
-                }
-            );
+
+        try {
+            const subscription = await fetchGoogleSubscription(subscriptionId, purchaseToken, packageName);
+
+            if (subscription) {
```
JamieB-gu: Can this be `undefined` as well as `null`?

typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -3,77 +3,63 @@ import {
...
+                const subscriptionExpiryDate: Date = new Date(parseInt(subscription.expiryTimeMillis));
```
JamieB-gu: `parseInt` can return `NaN` if the `string` passed to it cannot be parsed into an `Int`. When this happens, the expression `new Date(parseInt(subscription.expiryTimeMillis))` becomes a slightly surprising instance of an invalid `Date` where every property is `NaN`.

Under these circumstances, I think the condition `now > subscriptionExpiryDate` will *always* return `false` (as will `now <= subscriptionExpiryDate`, which is mathematically bizarre). The JSON-stringified response body will then become:

```json
{
  "subscriptionHasLapsed": false,
  "subscriptionExpiryDate": null
}
```

If this isn't the desired behaviour it may be worth adding a check for this scenario?

typescript/src/update-subs/google.ts
```diff
@@ -53,6 +23,25 @@ async function getGoogleSubResponse(record: SQSRecord): Promise<Subscription[]>
...
     }
+
+    if (!response) {
+        throw new ProcessingError("There was no data in google response", true);
+    }
+
+    const expiryDate = new Date(Number.parseInt(response.expiryTimeMillis));
+    return [new Subscription(
+        sub.purchaseToken,
+        new Date(Number.parseInt(response.startTimeMillis)).toISOString(),
```
JamieB-gu: Similar comment about `parseInt` and `NaN`.

typescript/src/subscription-status/googleSubStatus.ts
alexduf: I think you made some really good points about date parsing, I started to look in our codebase and that's bigger than an amendment on this PR. So I'll create a new PR after this series of PR :)

typescript/src/subscription-status/googleSubStatus.ts
alexduf: Fixed 👍

PR #143, opened 2020-02-21
Add trial period flag (part2)

alexduf:
In this PR I'm adding a `freeTrial` flag to both the subscription table and the event table.

For context, have a look at the [architecture diagram](https://docs.google.com/drawings/d/1Lt4cxTIjg_cSWd_bJCKpKwOZDFHTN7ki_GUZWJMBRvU/edit)

## Subscription table
Adding that flag to the subscription table is pretty easy as at the time we store the subscription, we have queried the subscription and the receipt from google and apple (respectively).

## Event table
Adding the flag to the event table is also pretty easy for Apple subscriptions, as Apple informs us of the state of that subscription in the payload we receive (I believe this is a recent addition).

That leaves us with the Google subscription events, which doesn't contain the trial information.
I had to refactor the code ([part 1](https://github.com/guardian/mobile-purchases/pull/142)) in order to make querying Google's service easier.

The primary goal of the PubSub lambda is to store the event and forward it to SQS for processing later. Any technical issue during that phase is critical as there's no guarantee neither Apple nor Google will retry to send the payload. (Although it is documented that both will retry).

As such, I made sure that querying that extra information would not cause the lambda to fail, no matter the outcome. A timeout, exception, temporary server outage on Google's side will all results in that flag being undefined in our database. I believe this is acceptable.

As I often do, this PR is easier to review commit per commit.

### DIFF
--- a/typescript/src/pubsub/google.ts
+++ b/typescript/src/pubsub/google.ts
@@ -1,11 +1,12 @@
 import 'source-map-support/register'
-import {ONE_YEAR_IN_SECONDS, parseStoreAndSend} from "./pubsub";
+import {parseStoreAndSend} from "./pubsub";
 import {SubscriptionEvent} from "../models/subscriptionEvent";
 import {dateToSecondTimestamp, thirtyMonths} from "../utils/dates";
 import {GoogleSubscriptionReference} from "../models/subscriptionReference";
 import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
 import {Option} from "../utils/option";
 import {fromGooglePackageName} from "../services/appToPlatform";
+import {fetchGoogleSubscription, GoogleResponseBody} from "../services/google-play";
 
 interface DeveloperNotification {
     version: string,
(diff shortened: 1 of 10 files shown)


davidfurey:
Is there a "not" missing from `As such, I made sure that querying that extra information would cause the lambda to fail,`?

typescript/src/pubsub/google.ts
```diff
@@ -46,7 +50,26 @@ const GOOGLE_SUBS_EVENT_TYPE: {[_: number]: string} = {
...
-export function toDynamoEvent(notification: DeveloperNotification): SubscriptionEvent {
+async function fetchMetadata(notification: DeveloperNotification): Promise<MetaData | undefined> {
+    try {
+        const subscription = await fetchGoogleSubscription(
+            notification.subscriptionNotification.subscriptionId,
+            notification.subscriptionNotification.purchaseToken,
+            notification.packageName
+        );
+        return {
+            freeTrial : subscription?.paymentState === 2 // 2 is freeTrial
```
davidfurey: Could you create an enum for PaymentState to avoid this magic number?

PR #147, opened 2020-02-24
Refactor the way we parse dates.

alexduf:
It should now be safer and clearer in case of a runtime error


typescript/src/services/appleValidateReceipts.ts
```diff
@@ -192,7 +198,7 @@ export function toSensiblePayloadFormat(response: AppleValidationServerResponse,
                 autoRenewStatus: autoRenewStatus,
                 cancellationDate: optionalMsToDate(receiptInfo.cancellation_date_ms),
                 expiresDate: new Date(expiryDate(receiptInfo)),
-                originalPurchaseDate: msToDate(receiptInfo.original_purchase_date_ms),
+                originalPurchaseDate: originalPurchaseDate ?? undefined,
```
JamieB-gu: Does TypeScript error if you don't provide this check? It should be able to narrow the type based on the check against `null` you've provided, which is what I think is happening for `eventTime` in the `toDynamoEvent` above.

Does it work if you wrap the `return` in an `else`, or does that cause an error based on some code paths not returning properly? I don't know if it's able to determine whether a `throw` obviates the need for an explicit return.

alexduf: You're absolutely right, I had started with the `originalPurchaseDate ?? undefined` then added the test above it, then forgot to remove the undefined.

it does happily compile without it, so I'm removing it 👍

PR #149, opened 2020-02-26
Add the product billing

alexduf:
Added the mapping between product id and the billing period.

It's non critical data, so if it's missing it won't be an issue.

I did add a log line to warn us if we spot a new product id, although we'd have to look at the logs to learn about it. I'm in two minds between leaving it like that and going all the way to sending a production alarm when we spot a new product id.


alexduf:
Tested locally

PR #152, opened 2020-03-24
Change the definition of "free trial" for iOS subs

alexduf:
It seems like the Guardian and Google agree on what a "free trial" is, but Apple has a slightly more nuanced definition.

This fixes the flag for all new events and all subscriptions that will be updated in the future.

Aside from this change I'm preparing a script that will update our existing database rows with the same definition.

### DIFF
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -69,7 +69,8 @@ export function toDynamoEvent(notification: StatusUpdateNotification): Subscript
         console.warn(`Unknown bundle id ${receiptInfo.bid}`)
     }
 
-    const freeTrial = receiptInfo.is_trial_period === "true";
+    // The Guardian's "free trial" period definition is slightly different from Apple, hence why we test for is_in_intro_offer_period
+    const freeTrial = receiptInfo.is_trial_period === "true" || receiptInfo.is_in_intro_offer_period === "true";
 
     return new SubscriptionEvent(
         receiptInfo.original_transaction_id,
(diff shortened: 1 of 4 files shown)


typescript/src/pubsub/apple.ts
```diff
@@ -69,7 +69,8 @@ export function toDynamoEvent(notification: StatusUpdateNotification): Subscript
         console.warn(`Unknown bundle id ${receiptInfo.bid}`)
     }
 
-    const freeTrial = receiptInfo.is_trial_period === "true";
+    // The Guardian's "free trial" period definition is slightly different from Apple, hence why we test for is_in_intro_offer_period
+    const freeTrial = receiptInfo.is_trial_period === "true" || receiptInfo.is_in_intro_offer_period === "true";
```
webb04: Any way these can be booleans?

typescript/tests/services/appleValidateReceipts.test.ts
```diff
@@ -11,6 +11,7 @@ describe("The apple validation service", () => {
             latest_expired_receipt_info: {
```
webb04: Worth adding a new test?

alexduf: Probably! Let me do that

alexduf: done :)

typescript/src/pubsub/apple.ts
alexduf: Nope, we're talking about Apple which like to "think different"

webb04: 🍏🔥

PR #154, opened 2020-04-07
Bump Google OAuth To 0.20.0

JamieB-gu:
## Why?

Bringing this library up to date.


jacobwinch:
I noticed a build error after approving this 🤦‍♂ If I understand the error correctly, [this Stack Overflow post](https://stackoverflow.com/a/55557287) looks helpful.

PR #155, opened 2020-04-30
Improve GooglePlaySubsStatus5xxErrors alert and silence CODE alerts

jacobwinch:
#### GooglePlaySubsStatus5xxErrors Improvements

https://theguardian.atlassian.net/browse/MSS-1603

- We now alert on % of requests which resulted in 5XX errors, instead of the number of 5XXs. This means that minor issues (e.g. transient problems with the Play Developer API) should not result in an alert. We'll still be informed about more serious problems, which is useful because we will need to escalate these.
- I've also added a runbook link to help with debugging. 

#### Silencing CODE alerts

I accidentally notified everyone when testing this change. I think that firing alerts related to CODE environments is unhelpful because it makes us less sensitive to real problems. To prevent this from happening again, I've set `ActionsEnabled` to `FALSE` in CODE. Alerts can still be tested to some extent (by looking at the CloudWatch UI) but the rest of the team are not impacted/notified.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -721,16 +721,16 @@ Resources:
           MetricStat:
             Metric:
               MetricName: Count
-                Namespace: AWS/ApiGateway
-                Dimensions:
-                  - Name: ApiName
-                    Value: !Sub ${App}-${Stage}
-                  - Name: Method
-                    Value: GET
-                  - Name: Resource
-                    Value: /google/subscription/{subscriptionId}/status
-                  - Name: Stage
-                    Value: !Ref Stage
+              Namespace: AWS/ApiGateway
+              Dimensions:
+                - Name: ApiName
+                  Value: !Sub ${App}-${Stage}
+                - Name: Method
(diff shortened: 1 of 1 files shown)


jacobwinch:
I've confirmed that this can [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/57ec2456-c271-4413-a805-1d58c178d0e7).

cloudformation.yaml
```diff
@@ -690,28 +698,56 @@ Resources:
...
-        - Name: Resource
-          Value: /google/subscription/{subscriptionId}/status
-        - Name: Stage
-          Value: !Ref Stage
       EvaluationPeriods: 1
-      MetricName: 5XXError
-      Namespace: AWS/ApiGateway
-      Period: 300
-      Statistic: Sum
       Threshold: 20
```
jacobwinch, line 750: I'm not too opinionated on this threshold - let me know if you have any thoughts.

PR #157, opened 2020-05-22
Better error messages

alexduf:
What we currently have in prod is a bit hard to trace


alexduf:
I think the 14 day retention works around the issue, so it doesn't really matter

alexduf:
it has you're right, but it was because we had concerns about putting these in the logs so I'll close it I think

PR #158, opened 2020-06-11
MSS-1648: Amend the link-purchases lambda to accept the package-name

frankie297:
## What does this change?
This allows us to pass package name to the platform

It has been tested on CODE

Worked on with @dskamiotis


davidfurey:
Thanks Frankie!

PR #166, opened 2020-10-26
Handle Identity 403

JamieB-gu:
## Why?

@jacobwinch and I discovered that we're logging a consistent stream of 5XX errors in the `google-link-user-subscription` lambda, which resolve to `403`/"Access Denied" messages in the logs. We've traced this back to calls to the Identity API, which we believe is returning `403`s when provided with invalid tokens from the Android app.

We'd have expected this to result in `mobile-purchases` returning `401` or `403` errors, but these were being upgraded to `5XX`. This is because the [REST client used](https://github.com/microsoft/typed-rest-client#rest) to contact Identity `throw`s on `4XX`s (except `404`s), which meant that it was bypassing [the handling logic](https://github.com/guardian/mobile-purchases/blob/6aea7549f6ad9e2256c02b54e3632832490a0e61/typescript/src/link/link.ts#L80) and breaking out [to the catch-all](https://github.com/guardian/mobile-purchases/blob/6aea7549f6ad9e2256c02b54e3632832490a0e61/typescript/src/link/link.ts#L85) that returns `500`.

We've introduced an additional `try...catch` in the function that calls identity, to handle the `403` case. Any other errors will be re-thrown and handled as before.

**Note:** The `getUserId` method is used in [one other place](https://github.com/guardian/mobile-purchases/blob/6aea7549f6ad9e2256c02b54e3632832490a0e61/typescript/src/user/user.ts#L96) (that I can find) - it looks like the same issue may crop up here (the upgrade to `500`), so I don't think our change should cause any problems.

## Changes

- Added additional `try...catch` to `getUserId` to deal with the REST API `throw`ing on `403`s

### DIFF
--- a/typescript/src/utils/guIdentityApi.ts
+++ b/typescript/src/utils/guIdentityApi.ts
@@ -19,11 +19,23 @@ export async function getUserId(headers: HttpRequestHeaders): Promise<Option<str
     const url = "https://id.guardianapis.com/user/me";
     const identityToken = getAuthToken(headers);
 
-    const response = await restClient.get<IdentityResponse>(url, {additionalHeaders: {Authorization: `Bearer ${identityToken}`}})
+    try {
+        const response = await restClient.get<IdentityResponse>(url, {additionalHeaders: {Authorization: `Bearer ${identityToken}`}})
 
-    if(response.result) {
-        return response.result.user.id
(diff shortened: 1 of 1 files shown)


alexduf (approved):
Looks great to me 👍

PR #167, opened 2020-11-25
Use shorter evaluation period for DLQ alarms

jacobwinch:
There is currently a 50 minute lag between messages hitting the DLQ and an alarm being received e.g. if the threshold is crossed at 02:35, the alarm doesn't fire until 03:25:

image

This seems to happen because we are currently using an evaluation period of 3000 seconds (or 50 minutes). This delay is undesirable and causes confusion when debugging. Shortening the evaluation period to 1 minute should allow us to learn about these problems more quickly. We use this same approach for other DLQ alarms [e.g. the Harvester alarm](https://github.com/guardian/mobile-n10n/blob/master/notificationworkerlambda/harvester-cfn.yaml#L197-L213).

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -857,7 +857,7 @@ Resources:
       Dimensions:
         - Name: QueueName
           Value: !GetAtt 'AppleSubscriptionsQueueDlq.QueueName'
-      Period: 3000
+      Period: 60
       Statistic: Sum
       EvaluationPeriods: 1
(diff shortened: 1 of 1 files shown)


jacobwinch:
I've confirmed that this CFN change is valid this by [deploying to `CODE`](https://riffraff.gutools.co.uk/deployment/view/6d512080-1478-48f5-b9bd-f5ddb889514e).

frankie297 (approved):

PR #168, opened 2020-12-10
Add some payload and receipt logs

frankie297:
## What does this change?
Add some logs to debug payload errors and invalid receipts easier.

This PR really is around hoping to gain some clarity on an error currently in our logs, and I want to see what the payload and receipt returns when we get this error.

Screenshot 2020-12-10 at 10 28 30

### DIFF
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -51,7 +51,6 @@ export function parsePayload(body: Option<string>): Error | StatusUpdateNotifica
     try {
         let notification = JSON.parse(body ?? "") as StatusUpdateNotification;
         delete notification.password; // no need to keep that in memory
-        console.log(`Notification payload: ${notification}`)
         return notification;
     } catch (e) {
         console.log("Error during the parsing of the HTTP Payload body: " + e);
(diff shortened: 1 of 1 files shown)


JamieB-gu (commented):
Code looks good.

Is there any PII in the receipt or notification payloads that we want to avoid logging out?

frankie297:
> Code looks good.
> 
> Is there any PII in the receipt or notification payloads that we want to avoid logging out?

      delete notification.password; // no need to keep that in memory
      console.log(`Notification payload: ${notification}`)

I was presuming that because I was using console.log after we delete the password we should be fine there. Would that not be the case? That's the only bit of sensitive info i can see really.

JamieB-gu:
Can we restrict the logging to information that we're confident are fine to log out? For example, the above error you included in the description suggests to me that either `latest_receipt_info` or `latest_expired_receipt_info` are `undefined`. So we _could_ do:

```ts
console.log(`latest_receipt_info is undefined: ${latest_receipt_info === undefined}, latest_expired_receipt_info is undefined: ${latest_expired_receipt_info === undefined}`);
```

jacobwinch (approved):

PR #169, opened 2020-12-10
Add alarm for Apple Pubsub API gateway, to alert us for 5XX errors

frankie297:
## What does this change?
Pretty much lifted this from @jacobwinch previous PR adding the same for Google Play: #155 

This should alert us to a spike in 5XX errors so we can deal with them more efficiently

NB: There is a current ongoing investigation into 5XX errors on this service so am inclined to merge this PR when that has been fixed as it may create some noise

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -826,7 +826,7 @@ Resources:
                 - Name: ApiName
                   Value: !Sub ${App}-${Stage}
                 - Name: Method
-                  Value: GET
+                  Value: POST
                 - Name: Resource
                   Value: /apple/subscription/status
                 - Name: Stage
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -783,6 +783,62 @@ Resources:
...
+          Label: Total number of requests
+          MetricStat:
+            Metric:
+              MetricName: Count
+              Namespace: AWS/ApiGateway
+              Dimensions:
+                - Name: ApiName
+                  Value: !Sub ${App}-${Stage}
+                - Name: Method
+                  Value: GET
```
jacobwinch, line 829:
```suggestion
                  Value: POST
```

cloudformation.yaml
```diff
@@ -783,6 +783,89 @@ Resources:
...
+                - Name: Resource
+                  Value: /apple/subscription/status
+                - Name: Stage
+                  Value: !Ref Stage
+            Period: 600
+            Stat: Sum
+          ReturnData: false
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: 1
+      Threshold: 20
```
jacobwinch, line 839: 20% 5XX errors is a high threshold to tolerate. We had to be much more tolerant than I'd usually like to be with the Google endpoint, because we often saw errors from their service and the alarm became very noisy for problems that we were unable to  action. 

We should have a look at the CloudWatch graphs for the Apple version and see if we can reduce this threshold.

cloudformation.yaml
```diff
@@ -783,6 +783,89 @@ Resources:
...
+          Value: /apple/pubsub
+        - Name: Stage
+          Value: !Ref Stage
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: 1
+      MetricName: 5XXError
+      Namespace: AWS/ApiGateway
+      Period: 300
+      Statistic: Sum
+      Threshold: 1
```
jacobwinch, line 866: An absence of errors is a good thing in this case so I think we should be [treating missing data](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-treatmissingdata) as `notBreaching`.

```suggestion
      Threshold: 1
      TreatMissingData: notBreaching
```

jacobwinch (approved):

jacobwinch:
The final changes look good to me 🚢 👍

PR #177, opened 2021-01-07
Create .nvmrc

JamieB-gu:
## Why?

Even if we're not explicitly using `nvm` in this project, we use it widely across the dept. Including an `.nvmrc` is a hint to all devs on this project for what Node version they should be using.

FYI @mxdvl

## Changes

- Specified Node 10 in new `.nvmrc`


marjisound (approved):

mxdvl (changes requested):
Better to pin to a patch version!

.nvmrc
```diff
@@ -0,0 +1 @@
+10
```
mxdvl, line 1:
```suggestion
10.22.1
```

frankie297 (approved):

JamieB-gu:
> Better to pin to a patch version!

Can do, although the lambda runtime doesn't allow you to specify: it's just `nodejs10.x`.

PR #178, opened 2021-01-07
LIVE-1234: Add unified receipt

frankie297:
## What does this change?
Whilst investing an iOS issue around missing subscriptions, we noticed an increased amount of errors in the logs of the `mobile-purchases-applepubsub` lambda. Having dug a little we could see an error message everytime we called the Sandbox environment:

Screenshot 2020-12-10 at 17 09 21

We could also see consistently 5XX error requests  from the Apple Pubsub API gateway over a period of time:

Screenshot 2020-12-10 at 10 48 55

Checking Apple documentation, they have decided to deprecate some of the fields from the [response body](https://developer.apple.com/documentation/appstoreservernotifications/responsebody):

Screenshot 2021-01-08 at 15 31 19

These fields have already been deprecated in the Sandbox environment and so thought we was missing information that now lives in a new `unified-receipt` object.

This PR adds the new object, and prepares for the deprecation of the fields in March.

## How to test
We tested this by:
-  Deploying our branch to CODE
- In API gateway running a test event on the POST request to the pubsub endpoint in CODE, passing in some test data as a payload
- Making sure we receive a 200 response

We tested different scenarios:
- If the unified-receipt and the deprecated fields are both sent (this is current behaviour in PROD)
- If the unified-receipt is missing and the deprecated fields remain
- If the deprecated fields are missing and unified-receipt is there

It is worth noting that currently in our production data we are receiving the unified-receipt already as Apple have rolled it out in production. So our payload consists of the to be deprecated fields, and the unfied-receipt.

We considered leaving the deprecated fields in the code base until March and removing them then, but we aren't going to be using them anymore so decided this might not be a problem.

We have also added a PR to add some metrics around alarming us for 5XX errors, which will be merged following a merge of this PR: https://github.com/guardian/mobile-purchases/pull/169

Paired with @marjisound @dskamiotis @jacobwinch

### DIFF
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -6,6 +6,7 @@ import {AppleSubscriptionReference} from "../models/subscriptionReference";
 import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
 import {Option} from "../utils/option";
 import {fromAppleBundle} from "../services/appToPlatform";
+import {PendingRenewalInfo} from "../services/appleValidateReceipts";
 
 // this is the definition of a receipt as received by the server to server notification system.
 // Not to be confused with apple's receipt validation receipt info (although they do look similar, they are different)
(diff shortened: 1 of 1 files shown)


alexduf (approved):
I'm glad apple is addressing their own complexity at long last!

PR #185, opened 2021-01-12
LIVE-1234: Add unified receipt, make sure latest_receipt is list type

frankie297:
## What does this change?
The original PR is here:https://github.com/guardian/mobile-purchases/pull/178

When this went into production we noticed some log errors which didn't surface during our testing. `latest_receipt_info` is now an array/list rather than a single value.

This PR fixes that issue, and also adds a jest test to make sure we catch this is the future should it ever change.

We have once again tested using API gateway test too and received a 200 response.

I will also monitor the logs after merging.

paired with @marjisound

### DIFF
--- a/typescript/src/export/exportEvents.ts
+++ b/typescript/src/export/exportEvents.ts
@@ -9,7 +9,7 @@ import {plusDays} from "../utils/dates";
 function cleanupEvent(subEvent: SubscriptionEvent): any {
     if (subEvent.applePayload) {
         delete subEvent.applePayload.password; // just to be safe
-        delete subEvent.applePayload.latest_receipt; // really no need to put that in the datalake
+        delete subEvent.applePayload.unified_receipt.latest_receipt; // really no need to put that in the datalake
     }
     return subEvent;
 }
--- a/typescript/src/export/exportEvents.ts
+++ b/typescript/src/export/exportEvents.ts
@@ -9,6 +9,7 @@ import {plusDays} from "../utils/dates";
 function cleanupEvent(subEvent: SubscriptionEvent): any {
     if (subEvent.applePayload) {
         delete subEvent.applePayload.password; // just to be safe
+        delete subEvent.applePayload.latest_receipt;
         delete subEvent.applePayload.unified_receipt.latest_receipt; // really no need to put that in the datalake
     }
     return subEvent;
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -34,7 +34,7 @@ export interface AppleReceiptInfo {
 export interface UnifiedReceiptInfo {
     environment: string,
     latest_receipt: string,
-    latest_receipt_info: AppleReceiptInfo,
+    latest_receipt_info: AppleReceiptInfo[],
     pending_renewal_info: PendingRenewalInfo[],
     status: number
 }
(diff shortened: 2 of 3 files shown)


alexduf:
Oh that rings a bell, I think I've been bitten by that before. Check that it's an array _all the time_ because it's possible it's an array _sometimes_.

And yes I realise it sounds bonkers to type a property as being either `A`, or `A[]`, but that's what I've seen in prod (of course it's not documented!)

I had to add special support here
https://github.com/guardian/mobile-purchases/blob/master/typescript/src/services/appleValidateReceipts.ts#L45-L46

And here
https://github.com/guardian/mobile-purchases/blob/master/typescript/src/services/appleValidateReceipts.ts#L137-L138

frankie297:
> Oh that rings a bell, I think I've been bitten by that before. Check that it's an array _all the time_ because it's possible it's an array _sometimes_.
> 
> And yes I realise it sounds bonkers to type a property as being either `A`, or `A[]`, but that's what I've seen in prod (of course it's not documented!)
> 
> I had to add special support here
> https://github.com/guardian/mobile-purchases/blob/master/typescript/src/services/appleValidateReceipts.ts#L45-L46
> 
> And here
> https://github.com/guardian/mobile-purchases/blob/master/typescript/src/services/appleValidateReceipts.ts#L137-L138

Yeh I saw this, and thought it might be the same, but I think apple have changed the structure of a `latest_receipt`

alexduf:
From my experience with them I wouldn't assume the first receipt is the latest! Maybe sort the receipts by expiry date first?

frankie297:
> From my experience with them I wouldn't assume the first receipt is the latest! Maybe sort the receipts by expiry date first?

Ha true, I feel like they are a bit rogue. Good idea 👍

frankie297:
@alexduf I added the `expires_date_ms` field to the payload so I can sort by unix epoch timestamp. Tested in CODE, confirmed the sorting works so we can then use the first receipt in the list to check for free trials. 

Also made sure that we were deleting the right` latest_receipt` before exporting to the datalake.

The thing I am unsure of, is how all this will affect the legacy part of the codebase, if I need to worry, and if I need to change this too.

typescript/src/export/exportEvents.ts
```diff
@@ -9,7 +9,7 @@ import {plusDays} from "../utils/dates";
 function cleanupEvent(subEvent: SubscriptionEvent): any {
     if (subEvent.applePayload) {
         delete subEvent.applePayload.password; // just to be safe
-        delete subEvent.applePayload.latest_receipt; // really no need to put that in the datalake
+        delete subEvent.applePayload.unified_receipt.latest_receipt; // really no need to put that in the datalake
```
alexduf, line 12: you probably want to still delete `subEvent.applePayload.latest_receipt` for the current production payloads

typescript/src/pubsub/apple.ts
```diff
@@ -62,25 +70,28 @@ export function parsePayload(body: Option<string>): Error | StatusUpdateNotifica
 export function toDynamoEvent(notification: StatusUpdateNotification): SubscriptionEvent {
     const now = new Date();
     const eventType = notification.notification_type;
-
-    const receiptInfo = notification.latest_receipt_info ?? notification.latest_expired_receipt_info;
-    console.log(`notification is from ${notification.environment}, latest_receipt_info is undefined: ${notification.latest_receipt_info === undefined}, latest_expired_receipt_info is undefined: ${notification.latest_expired_receipt_info === undefined}`);
-    const platform = fromAppleBundle(receiptInfo.bid);
+    const receiptInfo = notification.unified_receipt.latest_receipt_info;
```
alexduf, line 73: There's one more error case you might want to handle or log, which would be if the array is empty

typescript/tests/pubsub/pubsub.test.ts
```diff
@@ -99,3 +104,130 @@ describe("The google pubsub", () => {
...
+                        }
+                    ],
+                    status: 0
+                }
+            }
+        }
+
+        const expectedSubscriptionReferenceInSqs = {receipt: "TEST"};
+
+        return parseStoreAndSend(input, parseApplePayload, applePayloadToDynamo, toAppleSqsEvent, mockFetchMetadataFunction, mockStoreFunction, mockSqsFunction).then(result => {
```
alexduf, line 224: This test is a great addition 👍

alexduf (approved):

alexduf:
It feels like Apple if finally trying to clean up their messy API. There's hope for 2021 in the end.

typescript/src/pubsub/apple.ts
frankie297: ha, I had that and must have accidentally deleted it when I was clearing up after testing. 👍 thanks for pointing it out.

typescript/src/export/exportEvents.ts
frankie297: 👍

PR #193, opened 2021-01-20
Change Apple PubSub 5XX error threshold

frankie297:
## What does this change?
A threshold of 1 seems a bit too sensitive to alert us, 2 might seem more reasonable. If it triggers twice in 5 minutes then we know there is a problem.

I also noticed a typo in the alarm description.

@jacobwinch it would be good to get your opinion on this


marjisound (approved):

jacobwinch:
Seems sensible to me 👍

PR #197, opened 2021-01-22
handling error for apple 21007 response

marjisound:
## What does this change?

We noticed a spike of 5XX errors related to our newly merged work around the unified receipt, PR here: https://github.com/guardian/mobile-purchases/pull/185

The spike seems to coincide with the time I merged this piece of work in, seems like we need to make sure Apple can revalidate our receipts correctly.

Jira: [https://theguardian.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=LIVE&modal=detail&selectedIssue=LIVE-1458&assignee=5fd76101849d6401116426a3](url)

## How to test
1- **Test in CODE env**, we can test the [POST] apple/pubsub api endpoint in CODE env and make sure the whole flow is taken successfully for the request. 

2- **To test the impact in PROD**, we should create a subscription for a test user in apple device 
Then apple should call our [POST] apple/pubsub api endpoint which will trigger lambda **mobile-purchases-applepubsub-PROD**
Then a message is published to **mobile-purchases-PROD-apple-subscriptions-to-fetch** SQS queue 
The message will be processed by the **mobile-purchases-apple-update-subscriptions-PROD** lambda which will call apple to validate receipt. 
The result of this Lambda call should be successful and only a warning message should be displayed in the lambda logs. 

The expectation is that the number of 500 InternalServerErrors should go down in the **mobile-purchases-PROD** metrics.

### DIFF
--- a/typescript/src/models/GracefulProcessingError.ts
+++ b/typescript/src/models/GracefulProcessingError.ts
@@ -1,10 +1,8 @@
 export class GracefulProcessingError extends Error {
     message: string;
-    shouldRetry?: boolean;
 
     constructor(message: string, shouldRetry?: boolean) {
         super(message);
         this.message = message;
-        this.shouldRetry = shouldRetry;
     }
 }
--- a/typescript/src/models/GracefulProcessingError.ts
+++ b/typescript/src/models/GracefulProcessingError.ts
@@ -1,0 +1,10 @@
+export class GracefulProcessingError extends Error {
+    message: string;
+    shouldRetry?: boolean;
+
+    constructor(message: string, shouldRetry?: boolean) {
+        super(message);
+        this.message = message;
+        this.shouldRetry = shouldRetry;
+    }
+}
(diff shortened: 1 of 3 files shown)


typescript/src/models/GracefulProcessingError.ts
```diff
@@ -0,0 +1,10 @@
+export class GracefulProcessingError extends Error {
+    message: string;
+    shouldRetry?: boolean;
```
frankie297, line 3: I wonder if we need the retry?

frankie297 (approved):
👍 I approve but it would be good to get the opinion of someone who didn't pair with you on this 😄

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -104,7 +105,7 @@ function checkResponseStatus(response: AppleValidationServerResponse): AppleVali
         const msg = `Got status 21007 and we're in ${Stage}, so we are processing a receipt from the wrong environment. ` +
             `This shouldn't have happen as we should already retry receipts in sandbox if the return code was 21007`;
```
alexduf, line 106: I guess this log line isn't true as we it can happen if the `validateReceipt` function is called without the retry option. I suppose the whole line can be dropped?

alexduf (approved):

PR #199, opened 2021-01-26
aws lambda request header model changed

marjisound:
## What does this change?
Dependabot made a PR for aws-lambda version bump [(https://github.com/guardian/mobile-purchases/pull/194)](url)  from 8.10.27 to 8.10.71 but the new version was causing the build to fail due to the request header type that had been changed in APIGatewayProxyEvent from **[name: string]: string** to **[name: string]: string | undefined** 

This change was tested in Code env. 

## How to test
call to /apple/linkToSubscriptions - POST is an endpoint that is involved in the change. It was tested in CODE and the header was successfully fetched.

As long as the build is successful and the first call to /apple/linkToSubscriptions - POST is successful we can be sure that the change works fine.

### DIFF
--- a/typescript/src/services/google-play.ts
+++ b/typescript/src/services/google-play.ts
@@ -37,7 +37,7 @@ export function getAccessToken(params: S3.Types.GetObjectRequest) : Promise<Acce
         })
 }
 
-export function buildGoogleUrl(subscriptionId: string, purchaseToken: string, packageName: string) {
+export function buildGoogleUrl(subscriptionId: string | undefined, purchaseToken: string, packageName: string) {
     const baseUrl = `https://www.googleapis.com/androidpublisher/v3/applications/${packageName}/purchases/subscriptions`;
     return `${baseUrl}/${subscriptionId}/tokens/${purchaseToken}`;
 }
@@ -50,7 +50,7 @@ export interface GoogleResponseBody {
     paymentState: 0 | 1 | 2 | 3
 }
 
-export async function fetchGoogleSubscription(subscriptionId: string, purchaseToken: string, packageName: string): Promise<GoogleResponseBody | null> {
+export async function fetchGoogleSubscription(subscriptionId: string | undefined, purchaseToken: string, packageName: string): Promise<GoogleResponseBody | null> {
     const url = buildGoogleUrl(subscriptionId, purchaseToken, packageName);
     const accessToken = await getAccessToken(getParams(Stage));
     const response = await restClient.get<GoogleResponseBody>(url, {additionalHeaders: {Authorization: `Bearer ${accessToken.token}`}});
--- a/typescript/src/services/google-play.ts
+++ b/typescript/src/services/google-play.ts
@@ -37,8 +37,7 @@ export function getAccessToken(params: S3.Types.GetObjectRequest) : Promise<Acce
         })
 }
 
-export function buildGoogleUrl(subscriptionId: string | undefined, purchaseToken: string, packageName: string) {
-    if(subscriptionId == undefined) throw Error("subscriptionId is undefined");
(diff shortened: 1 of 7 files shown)


dskamiotis (approved):

frankie297 (approved):
Great work 🎉

typescript/src/services/google-play.ts
```diff
@@ -37,7 +37,7 @@ export function getAccessToken(params: S3.Types.GetObjectRequest) : Promise<Acce
         })
 }
 
-export function buildGoogleUrl(subscriptionId: string, purchaseToken: string, packageName: string) {
+export function buildGoogleUrl(subscriptionId: string | undefined, purchaseToken: string, packageName: string) {
     const baseUrl = `https://www.googleapis.com/androidpublisher/v3/applications/${packageName}/purchases/subscriptions`;
     return `${baseUrl}/${subscriptionId}/tokens/${purchaseToken}`;
```
JamieB-gu, line 42: If `subscriptionId` _is_ `undefined` then I think this will return:

```https://www.googleapis.com/androidpublisher/v3/applications/packageName/myPackageName/subscriptions/undefined/tokens/myToken
```

I suspect we may want to throw/return an error in that case rather than trying to make a request to that URL?

marjisound: Does it look better now?

typescript/src/services/google-play.ts
```diff
@@ -37,7 +37,8 @@ export function getAccessToken(params: S3.Types.GetObjectRequest) : Promise<Acce
         })
 }
 
-export function buildGoogleUrl(subscriptionId: string, purchaseToken: string, packageName: string) {
+export function buildGoogleUrl(subscriptionId: string | undefined, purchaseToken: string, packageName: string) {
+    if(subscriptionId == undefined) throw Error("subscriptionId is undefined");
```
JamieB-gu, line 41: Always use [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality) in JS unless there's a specific reason to use [standard equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality). It's also [good practice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else#description) to use block statements when working with `if`.

```suggestion
    if (subscriptionId === undefined) { throw Error("subscriptionId is undefined") };
```

Regarding throwing an error here, that seems to be the pattern used elsewhere in this repo (e.g. the `getAccessToken` function further up this file does that). Unless @alexduf or @frankie297 see any problems doing that, LGTM.

marjisound: Thanks @JamieB-gu for your comment. I addressed the suggestions you made. Also changed the place to throw the error in fetchGoogleSubscription. This function is used in 3 places in the code and in all of them, the general exception is caught.

typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -35,6 +35,9 @@ export async function handler(request: APIGatewayProxyEvent): Promise<APIGateway
         console.log(`Searching for valid ${subscriptionId} subscription for Android app with package name: ${packageName}`);
 
         try {
+            if (subscriptionId === undefined) {
```
alexduf, line 38: Hey that's much better, but now the exception thrown here will be caught line 57, when we could do some slightly nicer error handling and return with `HTTPResponses.INVALID_REQUEST` line 67.

To do so, I would probably implement a function very similar to `getPurchaseToken`, that gets the subscriptionId. Then line 32, I would only enter the condition if `subscriptionId` is set (no null, no undefined).

That way if subscriptionId is undefined we enter the else clause of the condition, and return `HTTPResponses.INVALID_REQUEST`

alexduf (approved):

PR #238, opened 2021-03-22
updated node to 14 in cloudformation

dskamiotis:
## What does this change?
We have some services on Node 10. These need updating by 2021-04-30, which is when Node 10 reaches EOL.

## How to test
By checking AWS Lambda service for mobile-purchases

We have tested in API Gateway with the Lambda and received a successful 2000 response.


.nvmrc
```diff
@@ -1 +1 @@
-10.22.1
+14.0.0
```
JamieB-gu, line 1: Latest version of 14 rather than `.0`?

dskamiotis: @JamieB-gu  - Node.js 14.16.1 | Fermium - apparently is the LTS version ensuring long term support so I went with this?

JamieB-gu (approved):

PR #259, opened 2021-04-19
LIVE:2308: Add stricter type checking for apple payload

frankie297:
## What does this change?
Adds stricter type checks for parsing the apple payload fields so we can easily debug errors when they occur, and find missing fields that should be included.

## How to test
We have tested the API gateway using a test payload.

The amount of errors we see decreases

Fields we were unsure of we have made optional, incase they are included sometimes but not all the time.

Paired with @JamieB-gu and @marjisound 

NB: In a future PR it would be good to add helper functions to clean up some of this code a bit.


groakland (approved):
Approved with some future suggestions detailed here: https://theguardian.atlassian.net/browse/LIVE-2364

PR #271, opened 2021-04-28
Live-1434 Updated to node 14 in cloudformation setup

dskamiotis:
## What does this change?
In order to complete the upgrade to Node 14, `cloudformation` export files are also updated


JamieB-gu (approved):
👍. It's worth checking what version the TeamCity builds are using for these and the previous changes too.

PR #274, opened 2021-04-29
Add more optional fields to apple receipt to debug parsing errors

frankie297:
## What does this change?
Add more optional fields to debug errors


shtukas:
Confirmed with Frankie that we do not need this anymore

PR #343, opened 2021-07-01
LIVE-2680: Remove sensitive token and receipt info from logs

frankie297:
## What does this change?
Prevents us logging out sensitive receipt and token information in error logs, which is a potential security risk.

## How to test
Send dummy data in a test environment and check that we no longer see the token information.

Security risks are reduced.


JamieB-gu (approved):

JamieB-gu:
FYI @stiliyana-simeonova @Reettaphant @groakland

stiliyana-simeonova:
👍 marvelous

PR #400, opened 2021-08-19
LIVE-2908 - create new lambda code for exporting subscription

marjisound:
## What does this change?
This PR is to create a new lambda for exporting the subscription data to datalake S3 bucket


NathanielBennett (approved):
:+1: Looks really good to me
Thanks!

PR #469, opened 2021-10-28
Test lambda script fixes

paulmr:
It wasn't possible to run the test lambda script because it wasn't being built when webpack was run, and webpack was needed to coalesce all of the dependencies into the correct ES version.

The README has also been updated with instructions on how to run it.

### DIFF
--- a/typescript/src/test-launcher/test-launcher.ts
+++ b/typescript/src/test-launcher/test-launcher.ts
@@ -1,7 +1,5 @@
 import * as fs from "fs";
 
-import path from 'path';
-
 if(process.argv.length == 4) {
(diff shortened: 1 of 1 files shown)


typescript/src/test-launcher/test-launcher.ts
```diff
@@ -1,8 +1,10 @@
 import * as fs from "fs";
 
+import path from 'path';
```
JamieB-gu, line 3: Is this used?

JamieB-gu (approved):

typescript/src/test-launcher/test-launcher.ts
paulmr: you are correct, it is not use. Thanks!

PR #481, opened 2021-11-08
Apple debug loggging

paulmr:
The aim of this PR is to provide detailed logging when an incoming notifiction from Apple fails to parse.

As we are unable to distinguish between production purchases, and 'sandbox' purchases done by readers through the Beta version of the app, we are filtering the logging through a process which fwill hide the values of the fields, and only print out the type of them. An exception is made for a list of fields which are OK to print, because we know they don't contain any personal information:

```typescript
const fieldAllowList = [ "environment", "product_id", "notification_type",
                         "auto_renew_status", "status", "purchase_date" ];
```

The output that is produced is constructed as a valid Json object, meaning that once you extract it and pretty print, you get something quite readable (this example is trimmed):

```json
{
  "auto_renew_product_id": "<string>",
  "bid": "<string>",
  "unified_receipt": {
    "pending_renewal_info": [
      {
        "original_transaction_id": "<string>",
        "auto_renew_status": "1",
        "product_id": "uk.co.guardian.gla.1month.2018May.withFreeTrial",
        "auto_renew_product_id": "<string>"
      }
    ],
    "latest_receipt_info": [ "... etc ..." ]
}

```

This is particluarly useful as have taken the output of this logging, and put it back into the lambda in the testing mode, to identify exactly which fields are missing.

The purpose of splitting out the type guards from one if statement:
```typescript
if(
        typeof payload.transaction_id === "string" &&
        typeof payload.product_id === "string" &&
        typeof payload.original_transaction_id === "string" && // ...
```

 into multiple:

```typescript
    if(typeof payload.transaction_id !== "string") return err("missing field: transaction_id")
    if(typeof payload.product_id !== "string") return err("missing field: product_id")
    if(typeof payload.original_transaction_id !== "string") return err("missing field: original_transaction_id")
    if(typeof payload.item_id !== "string" && typeof payload.item_id !== "undefined") return err(`incorrect optional field: item_id ${typeof payload.item_id}`)
```

is so that we can return a unique error message for each missing field, which is hugely helpful in debugging, but still maintains the type safety introduced by the type guards.

image

### DIFF
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -303,18 +303,19 @@ export function parsePayload(body: Option<string>): Error | StatusUpdateNotifica
     try {
         const notification: unknown = JSON.parse(body ?? "");
         if(isObject(notification) && notification?.environment === "Sandbox") {
-            let parseResultStr = "unknown"
-            try {
-                const parseResult = parseNotification(notification)
-                if(parseResult.kind === ResultKind.Err) {
-                    parseResultStr = parseResult.err
-                } else {
-                    parseResultStr = "ok"
-                }
-            } catch (e) {
-                parseResultStr = `exception: ${e}`
-            }
-            console.log(`debugLogPayload (parse result: ${parseResultStr}): ${JSON.stringify(debugLogPayload(notification))}`);
+            console.log(`debugLogPayload: ${JSON.stringify(debugLogPayload(notification))}`);
+            // let parseResultStr = "unknown"
+            // try {
+            //     const parseResult = parseNotification(notification)
+            //     if(parseResult.kind === ResultKind.Err) {
+            //         parseResultStr = parseResult.err
+            //     } else {
+            //         parseResultStr = "ok"
+            //     }
+            // } catch (e) {
+            //     parseResultStr = `exception: ${e}`
+            // }
+            // console.log(`debugLogPayload (parse result: ${parseResultStr}): ${JSON.stringify(debugLogPayload(notification))}`);
         }
         const parsedNotification = parseNotification(notification);
         if(parsedNotification.kind === ResultKind.Ok) {
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -302,10 +302,21 @@ function debugLogPayload(data: unknown, depth: number = 4, whitelisted: boolean 
 export function parsePayload(body: Option<string>): Error | StatusUpdateNotification {
     try {
         const notification: unknown = JSON.parse(body ?? "");
-        const parsedNotification = parseNotification(notification);
         if(isObject(notification) && notification?.environment === "Sandbox") {
-            console.log(`debugLogPayload (parse result: ${ResultKind[parsedNotification.kind]}): ${JSON.stringify(debugLogPayload(notification))}`);
+            let parseResultStr = "unknown"
+            try {
+                const parseResult = parseNotification(notification)
+                if(parseResult.kind === ResultKind.Err) {
+                    parseResultStr = parseResult.err
+                } else {
+                    parseResultStr = "ok"
+                }
+            } catch (e) {
+                parseResultStr = `exception: ${e}`
+            }
+            console.log(`debugLogPayload (parse result: ${parseResultStr}): ${JSON.stringify(debugLogPayload(notification))}`);
         }
+        const parsedNotification = parseNotification(notification);
         if(parsedNotification.kind === ResultKind.Ok) {
             return parsedNotification.value;
         }
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -86,7 +86,7 @@ const parseArray = <A>(parseA: (a: unknown) => Result<string, A>) => (array: unk
     return err("Is not an array");
 };
 
-const fieldWhiteList = [ "environment", "product_id", "notification_type",
+const fieldAllowList = [ "environment", "product_id", "notification_type",
                          "auto_renew_status", "status", "purchase_date" ];
 
 function debugCleanPayload(data: unknown, depth: number = 4, whitelisted: boolean = false): object | string {
@@ -99,7 +99,7 @@ function debugCleanPayload(data: unknown, depth: number = 4, whitelisted: boolea
         } else {
             let result: Record<string, unknown> = {}
             for(let k in data)
-                result[k] = debugCleanPayload(data[k], depth - 1, fieldWhiteList.includes(k))
+                result[k] = debugCleanPayload(data[k], depth - 1, fieldAllowList.includes(k))
             return result
         }
     } else if(whitelisted) return `${data}`
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -285,7 +285,7 @@ function parseNotification(payload: unknown): Result<string, StatusUpdateNotific
 
 const fieldWhiteList = [ "environment" ];
 
-function debugLogPayload(notification: unknown, depth: number = 1): string {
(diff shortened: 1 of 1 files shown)


typescript/src/pubsub/apple.ts
```diff
@@ -86,51 +86,73 @@ const parseArray = <A>(parseA: (a: unknown) => Result<string, A>) => (array: unk
...
+                res.push(debugCleanPayload(item, depth - 1))
+            return res
+        } else {
+            let result: Record<string, unknown> = {}
+            for(let k in data)
+                result[k] = debugCleanPayload(data[k], depth - 1, fieldAllowList.includes(k))
+            return result
+        }
+    } else if(whitelisted) return `${data}`
+    else return `<${typeof(data)}>`
```
marjisound, line 106: Could we not just do `else return typeof(data)` here?

typescript/src/pubsub/apple.ts
```diff
@@ -86,51 +86,73 @@ const parseArray = <A>(parseA: (a: unknown) => Result<string, A>) => (array: unk
...
+        purchase_date_ms: payload.purchase_date_ms,
+        original_purchase_date_ms: payload.original_purchase_date_ms,
+        expires_date: payload.expires_date,
+        expires_date_ms: payload.expires_date_ms,
+        is_in_intro_offer_period: payload.is_in_intro_offer_period,
+        is_trial_period: payload.is_trial_period,
+        bvrs: payload.bvrs,
+        version_external_identifier: payload.version_external_identifier
+    })
+//    return err(`Apple Receipt Info object from Apple cannot be parsed: ${debugLogPayload(payload)}`)
```
marjisound, line 155: here, if any of the fields is missing, the function returns error straight away and would not check the rest of the fields in case more was missing. I think it's fine but just wanted to see what you think about it?

marjisound (approved):
Great work @paulmr it's certainly looking more robust and cleaner. just left 2 questions though the PR looks great so feel free to ignore if you see fit

typescript/src/pubsub/apple.ts
paulmr: we could, but I thought visually it was a bit nicer to indicate that we were showing the type rather than the value, but it's purely visual and has no actual use so I am happy either way

typescript/src/pubsub/apple.ts
paulmr: yes, you make a very valid point, that is a limitation with this implementation. The problem is that if you don't return straight away after making the assertion, then it won't be a type guard any more, so I'm not sure how we could implement such a thing cleanly.

For example, if we converted [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwJYBMoF4oG8oDMD2+AXFAM7ABOSAdgOYA0UARnBSeVXVAL4DcAUP1wBXagGNgSfNShJS8ZCgAUxKACUIY-BRQAeDjQZRRAa2r4A7tQB8ASnaVDUAD6xEqbPyjfZuFRnRMamEEBBdXUEh8P3xbKABCQKgAInwmACtNYGS4ighgYQoZVNlSKHNgKDgZNMyJZIEfXyVIiGiofAA6AnwEpOSDOhyoPIKilIBbOVInXCQIBBQSHoavHyQ-Vvaulgo+zAHHIdz8wuKp0hmuOYWl5lZVptGz7DxCEi6exl2Pzt2eATcQRaaikfAICCdBD4WhKOQKVBKLDcWy2ARAA):

```typescript
type Valid = { foo: string, bar: string };

function isValid(o: Record<string, unknown>): string | Valid {
    if(o === null || typeof(o) !== "object") return "o is not an object";
    if(typeof o.foo !== "string") return "missing field: foo";
    if(typeof o.bar !== "string") return "missing field: bar";
    return { foo: o.foo, bar: o.bar };
}

console.log(isValid({}));
```

into [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwJYBMoF4oG8oDMD2+AXFAM7ABOSAdgOYA0UARnBSeVXVAL4DcAUP1wBXagGNgSfNShJS8ZCgAUxKACUIY-BRQAeDjQZRRAa2r4A7tQB8ASnaVDUAD6xEqbPyjeoCCMCgAWzlSQwAxJAgEFFIMKABtAF0BH1lcFQx0TGphBAQXV1BIfHT8WygAQiyoACJ8JgArTWAa8op-YQoZOtlY8wC4GXqmiRqUnyR0oogSqHwAOgJ8SuqagzpWoJDwyOjSebBhUgALJRql1vHvSaVp2YWWChXMNccN8uDSULoIqJiDo6nGqPS5eCbpT7fWi-PbzPx0YDHKDWKAABjaHS6tQAsts6CQalAANRbL47P77Br4GhnRig1LtYCdGQ4JYkBZLRiPdnzR48ATcQRaaikfB+OH4WhKOQKVBKLDcWy2ARAA):

```typescript
type Valid = { foo: string, bar: string };

function isValid(o: Record<string, unknown>): string | Valid {
    let missingFields = [];
    if(o === null || typeof(o) !== "object") return "o is not an object";
    if(typeof o.foo !== "string") missingFields.push("foo");
    if(typeof o.bar !== "string") missingFields.push("bar");
    if(missingFields.length > 0) return "Missing: " + missingFields.join(", ");
    return { foo: o.foo, bar: o.bar };
}

console.log(isValid({}));
```

Then it no longer compiles because we don't have an identifiable static assertion that in all cases when the code reaches the return statement, `o.foo` exists (even though we can see it does).

One [way](https://www.typescriptlang.org/play?ssl=15&ssc=26&pln=1&pc=1#code/C4TwDgpgBAaghgGwJYBMoF4oG8oDMD2+AXFAM7ABOSAdgOYA0UARnBSeVXVAL4DcAUP1wBXagGNgSfNShJS8ZCgAUxKACUIY-BRQAeDjQZRRAa2r4A7tQB8ASnaVDUAD6xEqbPyjeoCCMCgAWzlSQwAxJAgEFFIHTloAbQBdDChkgR88UQkpGWAIcgBJYBUSDS0dfUc6RlNzKzsSfFlSN0VPTMykXBUMdExqYQQEF1dQSHwe-FsoAEJ+qAAifCYAK01gRZmKf2EKGVxEUggMzu9upXGISah8ADoCZvnMRYM6LaCQ8Mjo0juwYSkAAWSkWjy2pzOFyuN3uLAocwWr2qtA+wVIoToESiMX+gJBi3hEK8ZygO2Aexk6MxtGxvzufjowCBfSgAAZITwSecevkiiVpttdvtsHhCE0HoRGPCJfCeJyoscycKZItqYYSIsoABqT4Y744v6rfA0UGMYncQRaaikfB+Bn4WhKOQKVBKLDcWy2XhAA) I can think of to do it is to use [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) but that feels a bit fiddly because:

* we have to create a separate function for each predicate
* as far as I can tell there is no way for them to return a value, so we would have to use side-effects to update some variable that was defined outside the scope of the function

e.g:

```typescript
type Valid = { foo: string, bar: string };

function isValid(o: Record<string, unknown>): string | Valid {
    let missingFields: string[] = [];
    function testIt(o: Record<string, unknown>): o is Valid {
        if(o === null || typeof(o) !== "object") return false;
        if(typeof o.foo !== "string") missingFields.push("foo");
        if(typeof o.bar !== "string") missingFields.push("bar");
        return missingFields.length == 0;
    }
    if(testIt(o)) return { foo: o.foo, bar: o.bar };
    else return "missing: " + missingFields.join(", ");
}

console.log(isValid({}));
```

This feels a bit like overkill to me, although it does provide a nicer result. What do you think?

typescript/src/pubsub/apple.ts
marjisound: Makes sense, I agree 👍

PR #483, opened 2021-11-09
deal with the fact that some fields marked as required aren't being sent

paulmr:
Apple seems to have changed something upstream in the fields that it is sending to us when a user makes a purchase through the App store.

We are seeing a number of failures in our logs:

image

After running through various iterations, fixing each field as it was reported missing in a test sample which I took out of the log message, I now have something that is parsed correctly.

The only meaningful change I had to make, other than to mark fields as optional, was in the sorting of the latest receipt. We are attempting to locate the latest entry by sorting by expiry, but we are no longer recieving this from Apple (at least much of the time we aren't). Instead we are going to use the purchase date, which we do have, and which seems likely to better reflect the intended use.

### DIFF
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -17,13 +17,13 @@ export interface AppleReceiptInfo {
     transaction_id: string,
     product_id: string,
     original_transaction_id: string,
-    web_order_line_item_id: string,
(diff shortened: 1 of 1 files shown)


marjisound (approved):
Looks good to me 👍

frankie297 (approved):
👍 looks great

PR #559, opened 2022-02-15
Ignore some validation exceptions

davidfurey:
## What does this change?

This will ignore validation failures as long as some receipts validate.

This might resolve some of the issues that Amuse labs have raised of users not being able to restore their subscriptions.


frankie297 (approved):
👍

PR #564, opened 2022-02-18
Ignore duplicate original transaction ids

davidfurey:
Filter out duplicate original transaction ids because dynamodb batch write doesn't allow you to write the same value for a range key multiple times in one transaction.

I don't know where the duplicates are coming from but this should mask the issue that we currently see:
`Internal Server Error ValidationException: Provided list of item keys contains duplicates`


JamieB-gu (commented):
Looks good from a TS perspective 👍

frankie297 (approved):

PR #584, opened 2022-03-10
Deduplicate link requests

davidfurey:
## What does this change?

De-duplicates receipts sent from iOS apps. This should hopefully fix the error `Internal Server Error ValidationException: Provided list of item keys contains duplicates`. 

We now filter out duplicate original transaction ids earlier in the process so that the dynamodb get doesn't include duplicates (see https://github.com/guardian/mobile-purchases/blob/master/typescript/src/link/link.ts#L19)

I think the live app does this client side (https://github.com/guardian/ios-live/blob/1f48fc91115ac5e6725017a9f8cc4c88d43fb33b/GLA/GLA/Classes/User%2BSubscriptionLink.swift#L82) but the puzzles app currently doesn't.

### DIFF
--- a/typescript/src/link/apple.ts
+++ b/typescript/src/link/apple.ts
@@ -17,7 +17,7 @@ type AppleLinkPayload = {
 function deduplicate<T, U>(list: T[], selector: (item: T) => U): T[] {
     return list.reduce<T[]>(
         (agg, item) =>
-            agg.some((x) => selector(x) == selector(item)) ?
+            agg.some((x) => selector(x) === selector(item)) ?
                 agg : agg.concat([item]), 
         []
     )
--- a/typescript/src/link/apple.ts
+++ b/typescript/src/link/apple.ts
@@ -14,16 +14,28 @@ type AppleLinkPayload = {
     subscriptions: AppleSubscription[]
 }
 
-function parseAppleLinkPayload(request: APIGatewayProxyEvent): AppleLinkPayload {
-    return JSON.parse(request.body ?? "") as AppleLinkPayload;
+function deduplicate<T, U>(list: T[], selector: (item: T) => U): T[] {
+    return list.reduce<T[]>(
+        (agg, item) =>
+            agg.some((x) => selector(x) == selector(item)) ?
+                agg : agg.concat([item]), 
+        []
+    )
+}
+
+export function parseAppleLinkPayload(request: APIGatewayProxyEvent): AppleLinkPayload {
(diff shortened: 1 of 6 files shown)


typescript/src/link/apple.ts
```diff
@@ -14,16 +14,28 @@ type AppleLinkPayload = {
     subscriptions: AppleSubscription[]
 }
 
-function parseAppleLinkPayload(request: APIGatewayProxyEvent): AppleLinkPayload {
-    return JSON.parse(request.body ?? "") as AppleLinkPayload;
+function deduplicate<T, U>(list: T[], selector: (item: T) => U): T[] {
+    return list.reduce<T[]>(
+        (agg, item) =>
+            agg.some((x) => selector(x) == selector(item)) ?
```
JamieB-gu, line 20: `===`?

JamieB-gu, line 20: `===`?

davidfurey: Thanks

frankie297 (approved):
👍

typescript/src/link/apple.ts
```diff
@@ -14,16 +14,28 @@ type AppleLinkPayload = {
...
+    return list.reduce<T[]>(
+        (agg, item) =>
+            agg.some((x) => selector(x) === selector(item)) ?
+                agg : agg.concat([item]), 
+        []
+    )
+}
+
+export function parseAppleLinkPayload(request: APIGatewayProxyEvent): AppleLinkPayload {
+    const parsed = JSON.parse(request.body ?? "") as AppleLinkPayload;
```
JamieB-gu, line 27: Just a heads up, `JSON.parse` can throw with a `SyntaxError`. You may or may not want to wrap this in a `try...catch` (or the `fromUnsafe` function from the `Result` type).

JamieB-gu (commented):
I suspect you're probably aware of this, but type casts (`as`), particularly to `any`s, come with a risk of runtime errors. They're basically overriding the type checker.

It doesn't look this project lints against this, so perhaps it's considered an acceptable trade-off here for convenience, but I thought I'd mention just to make sure you're intentionally making that trade-off.

typescript/src/link/apple.ts
davidfurey: Fair, although I'm not adding this potential exception or the cast, and I'm reluctant to do the refactoring required.  I expect the failure would cause an Internal server error when we should really return a Bad Request. I'm okay with that.

PR #591, opened 2022-03-15
Remove legacy scala code

frankie297:
## What does this change?
Removes the old legacy scala lambdas from the project, and any remaining scala dependencies

We may have a chance that the lambdas are still needed for super old devices, there may be a risk we need to revert this work.

Do we need the apps to force people to update to a newer version?


groakland (approved):
LGTM

frankie297:
Closing in favour of a new PR as we need to keep the googleoauth lambda

PR #595, opened 2022-03-17
Delete the legacy scala lambdas

frankie297:
## What does this change?
Deletes and cleans up the project, freeing us from the old scala lambas that have been hanging around. Streamlines our architecture.

Nothing breaks

That old devices might still be using this, and we might need to potentially revert.


paulmr (approved):
This looks great. We have done everything we can to verify that this code is not used any more. We will monitor closely after deploy to check if there are any problems.

waisingyiu (approved):
LGTM

PR #598, opened 2022-03-22
Update architecture diagram

waisingyiu:
## What does this change?

This PR is to update the architecture diagram according to the latest cloudformation definition.  The diagram also contains the actual name of AWS resources for easy reference.

## How to test

No change to any source code.  It only updates the architecture diagram in the image file which is shown on the README file.

N/A

Yes, no risks are expected.

## Images

No change to the program behaviour.

## Accessibility

N/A


frankie297 (approved):
Looks great, thanks Wai Sing!

PR #613, opened 2022-04-04
Make sure original_transaction_id is parsed correctly

frankie297:
## What does this change?
Analysts noticed that from the 25/03/22 numbers of subscriptions was dramatically dropping. When looking at the size of the files in the raw ophan bucket for mobile_subscriptions_events I could see that the file size was dramatically smaller, coinciding with the data the analysts saw.

When looking into the kibana logs, on 22/03/22 we started seeing parsing errors for the payload:

Screenshot 2022-04-04 at 17 13 30

When investigating it became clear that the `original_transaction_id`  for the `StatusUpdateNotification` payload was an optional field of `string`

In Apples documentations this however is now a `number`, we also suspect that once the field was optional for Apple but now they have changed it to be non-optional and so we fail to parse the payload.

This PR makes sure we still handle our current model of the type being a string, whilst also allowing us to accept number as a type too. 

In the future we should have better alarms in place to shout at us.

### DIFF
--- a/typescript/src/pubsub/apple.ts
+++ b/typescript/src/pubsub/apple.ts
@@ -273,7 +273,7 @@ function parseNotification(payload: unknown): Result<string, StatusUpdateNotific
         typeof payload.bid === "string" &&
         typeof payload.bvrs === "string" &&
         typeof payload.notification_type === "string" &&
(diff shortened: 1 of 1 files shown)


DavidLawes (approved):
Great find :) Out of interest, do we have a backlog of improvements we'd like to make for mobile-purchases (can we add alarming to that backlog, if it's not on there already)?

PR #684, opened 2022-06-28
Upgrade jest transitive dependencies to fix security vulnerability

frankie297:
## What does this change?
Upgrade jest transitive dependencies to fix security vulnerability


frankie297:
No longer needed

PR #820, opened 2022-12-22
Okta Integration

shtukas:
We upgrade the authentication mechanism used for the two mobile subscription links:

- https://mobile-purchases.mobile-aws.guardianapis.com/apple/linkToSubscriptions
- https://mobile-purchases.mobile-aws.guardianapis.com/google/linkToSubscriptions

to support the new Okta authentication. 

Note the two requirements:

1. The code should support both the old tokens and the new Okta token. 
2. In addition of returning 200 for valid requests and 401 for unauthorised requests, we now must also return 403 for forbidden requests corresponding to correct tokens but incorrect scopes. 

(2.) lead to the introduction of the type `UserIdResolution` which carries the information required to return the code code.

### DIFF
--- a/typescript/src/utils/guIdentityApi.ts
+++ b/typescript/src/utils/guIdentityApi.ts
@@ -18,7 +18,7 @@ interface OktaJwtVerifierReturn {
 }
 
 export interface UserIdResolution {
-    status: "incorrect-token" | "incorrect-scope" | "success",
+    status: "incorrect-token" | "incorrect-scope" | "missing-identity-id" | "success",
     userId: null | string
 }
 
@@ -68,28 +68,11 @@ async function getUserId_NewOkta(headers: HttpRequestHeaders): Promise<UserIdRes
             return await oktaJwtVerifier.verifyAccessToken(accessTokenString, expectedAud)
             .then((payload: OktaJwtVerifierReturn) => {
                 if (payload.claims.scp.includes(scope)) { 
-                    // We have found the email address claim
-                    // The claims attribute maps to
-                    /*
-                        {
-                            ver: 1,
-                            jti: 'EDITED',
-                            iss: 'https://profile.code.dev-theguardian.com/oauth2/aus3v9gla95Toj0EE0x7',
-                            aud: 'https://profile.code.dev-theguardian.com/',
-                            iat: 1672252869,
-                            exp: 1672256469,
-                            cid: '0oa4iyjx692Aj8SlZ0x7',
-                            uid: '00u38ar4186TSK9j00x7',
-                            scp: [ 'email', 'openid', 'profile' ],
-                            auth_time: 1671029934,
-                            sub: 'EDITED',
-                            identity_username: '',
-                            email_validated: true,
-                            legacy_identity_id: 'EDITED'
-                        }
-                    */
-                    // Let's use the legacy_identity_id
-                    return {status: "success", userId: payload.claims.legacy_identity_id};
+                    if (payload.claims.legacy_identity_id) {
+                        return {status: "success", userId: payload.claims.legacy_identity_id};
+                    } else {
+                        return {status: "missing-identity-id", userId: null};
+                    }
                 } else {
                     // We have passed authentication but we didn't pass the scope check
                     return {status: "incorrect-scope", userId: null};
@@ -116,7 +99,13 @@ async function getUserId_NewOkta(headers: HttpRequestHeaders): Promise<UserIdRes
     When we moved to the Okta authentication, we needed to make the difference between failures
     due to an incorrect token and failures due to incorrect scopes. In the case of an incorrect
     token we need to return { HTTPResponses.UNAUTHORISED, 401 } but in the case of incorrect scope
-    we need to return { HTTPResponses.FORBIDDEN, 403 }.
+    we need to return { HTTPResponses.FORBIDDEN, 403 }. Additionaly note that if the claim that carries the
+    identity id is missing, we return { HTTPResponses.INVALID_REQUEST, 400 }.
+
+    Incorrect token                     -> 401
+    Incorrect scope                     -> 403
+    Missing claim / Missing identity id -> 400
+    Successful request                  -> 200
 
     To be able to convey to { function: parseAndStoreLink } which case occured during the authentication,
     we are extending the return type of the getUserId functions, to become a { UserIdResolution }
--- a/typescript/src/utils/guIdentityApi.ts
+++ b/typescript/src/utils/guIdentityApi.ts
@@ -122,6 +122,13 @@ async function getUserId_NewOkta(headers: HttpRequestHeaders): Promise<UserIdRes
     we are extending the return type of the getUserId functions, to become a { UserIdResolution }
 */
 
+/*
+    Date: 07th Jan 2022
+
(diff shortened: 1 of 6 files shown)


shtukas:
Fully tested 

```
- old identity token               -> {"status": 200, "message": "OK"}
- invalid okta token               -> {"status": 401, "message": "UNAUTHORISED"}
- valid okta token (wrong scope)   -> {"status": 403, "message": "FORBIDDEN"}
- valid okta token (correct scope) -> {"status": 200, "message": "OK"}
```

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,120 @@ interface IdentityResponse {
...
+                            cid: '0oa4iyjx692Aj8SlZ0x7',
+                            uid: '00u38ar4186TSK9j00x7',
+                            scp: [ 'email', 'openid', 'profile' ],
+                            auth_time: 1671029934,
+                            sub: 'EDITED',
+                            identity_username: '',
+                            email_validated: true,
+                            legacy_identity_id: 'EDITED'
+                        }
+                    */
```
kelvin-chappell, line 105: Is this useful?

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,120 @@ interface IdentityResponse {
...
+                    // Let's use the legacy_identity_id
+                    return {status: "success", userId: payload.claims.legacy_identity_id};
```
kelvin-chappell, line 107: In other places we return a 400 if the claims we expect aren't in the payload.  ie in this case If there's no identity ID in the payload

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,120 @@ interface IdentityResponse {
...
+    In the case of null we would return a { HTTPResponses.UNAUTHORISED, 401 }
+
+    When we moved to the Okta authentication, we needed to make the difference between failures
+    due to an incorrect token and failures due to incorrect scopes. In the case of an incorrect
+    token we need to return { HTTPResponses.UNAUTHORISED, 401 } but in the case of incorrect scope
+    we need to return { HTTPResponses.FORBIDDEN, 403 }.
+
+    To be able to convey to { function: parseAndStoreLink } which case occured during the authentication,
+    we are extending the return type of the getUserId functions, to become a { UserIdResolution }
+*/
```
kelvin-chappell, line 34: This is useful note.  Could you also add what changes will be needed when we switch entirely to Okta?  It might be obvious so I'm not sure if it would help or not.

kelvin-chappell (commented):
Great work!  Just a few comments.

Would probably be good to get a review from mobile side as well.

typescript/src/utils/guIdentityApi.ts
shtukas: Done.

typescript/src/utils/guIdentityApi.ts
shtukas: Done.

typescript/src/utils/guIdentityApi.ts
shtukas: Removed.

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,120 @@ interface IdentityResponse {
...
+}
+
+export async function getUserId(headers: HttpRequestHeaders): Promise<UserIdResolution> {
+    // This function is the front that implement the common interface behind which 
+    // there is the old (Identity) and new (Okta) authentication methods.
+    const resolution = await getUserId_OldIdentity(headers);
+    if (resolution.status == "success") {
+        return resolution;
+    }
+    return await getUserId_NewOkta(headers);
```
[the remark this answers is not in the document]
shtukas: I am actually going to leave it that way, because the Okta authentification fails is more ways than the old identity authentication and in order to return the right code, it's better if the Okta authentication comes second.

shtukas:
At @kelvin-chappell 's request I expanded the list of return codes

```
- Valid old identity token        -> {"status": 200, "message": "OK"}

- Invalid okta token              -> {"status": 401, "message": "UNAUTHORISED"}

- Valid okta token & wrong scope  -> {"status": 403, "message": "FORBIDDEN"}

- Valid okta token & good scope & missing identity Id
                                  -> {"status": 400, "message": "INVALID_REQUEST"}

- Fully valid okta request        -> {"status": 200, "message": "OK"}
```

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,122 @@ interface IdentityResponse {
...
     }
 }
+
+async function getUserId_NewOkta(headers: HttpRequestHeaders): Promise<UserIdResolution> {
+    try {
+        const OktaJwtVerifier = require('@okta/jwt-verifier');
+
+        const ISSUER      = 'https://profile.code.dev-theguardian.com/oauth2/aus3v9gla95Toj0EE0x7'
+        const CLIENT_ID   = "0oa4iyjx692Aj8SlZ0x7"
+        const expectedAud = "https://profile.code.dev-theguardian.com/";
```
aoifemcl15, line 57: Is the plan to add a production equivalent of these configs at a later point, so that both the CODE and PROD environments will be supported?

aoifemcl15 (commented):
All looks good to me! 
I've just left one comment regarding the production configuration.

kelvin-chappell (approved):
👍

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,122 @@ interface IdentityResponse {
...
         }
     }
 }
+
+async function getUserId_NewOkta(headers: HttpRequestHeaders): Promise<UserIdResolution> {
+    try {
+        const OktaJwtVerifier = require('@okta/jwt-verifier');
+
+        const ISSUER      = 'https://profile.code.dev-theguardian.com/oauth2/aus3v9gla95Toj0EE0x7'
+        const CLIENT_ID   = "0oa4iyjx692Aj8SlZ0x7"
```
coldlink, line 56:
```suggestion
```

As mentioned in chat, clientId shouldn't be needed by an API which verifies access tokens

typescript/src/utils/guIdentityApi.ts
```diff
@@ -11,31 +10,122 @@ interface IdentityResponse {
...
+        const expectedAud = "https://profile.code.dev-theguardian.com/";
+        const scope       = "guardian.mobile-purchases-api.update.self"
+        
+        const oktaJwtVerifier = new OktaJwtVerifier({
+            issuer: ISSUER,
+            clientId: CLIENT_ID,
```
coldlink, line 62:
```suggestion
```

typescript/src/utils/guIdentityApi.ts
shtukas: @aoifemcl15 Thanks making the update now

typescript/src/utils/guIdentityApi.ts
shtukas: Thanks removing now :)

typescript/src/utils/guIdentityApi.ts
shtukas: 👍

PR #841, opened 2023-01-17
Upgrade @okta/jwt-verifier to version 3.0.0

shtukas:
## What does this change?

Bump the @okta/jwt-verifier version from 2.6.0 to 3.0.0. The main motivation is to avoid a vulnerability introduced in a transitive dependancy (jsonwebtoken). This has been fixed in 3.0.0 

Although marked as a major version change, the change only affects the required version of node, which has now been pushed to 14. This required an update of the TeamCity config for the build stage. And we also checked that the lambdas are using the right version of node on AWS

Screenshot 2023-01-18 at 13 35 01

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@aws/dynamodb-data-mapper-annotations": "^0.7.3",
     "@aws/dynamodb-expressions": "^0.7.3",
     "@guardian/types": "^6.0.0",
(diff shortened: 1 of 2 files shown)


shtukas:
Tested on CODE

DavidLawes (approved):
👍

PR #889, opened 2023-03-01
Implement Soft Opt In version 1

shtukas:
As part of the Soft Opt In project, and notably what is referred to as version 1 (or stage 1 of a multi stage migration) our friends at Retention requested that when users ('s mobile phones) hit the two endpoints:

- `/google/linkToSubscriptions` and 
- `/apple/LinkToSubscriptions`

then that a consent notification is sent to the Identity API. The payload is a constant object, and the only variation is the user Id, the identity Id, which we extract from the user having had to authenticate to the end point. So it was an easy change.

Note: we set  `node-fetch` as new dependency. Note that we already had `typed-rest-client` able to make HTTP requests, but I am not aware of it being able to make PATCH requests.

### DIFF
--- a/typescript/src/link/apple-utils.ts
+++ b/typescript/src/link/apple-utils.ts
@@ -1,0 +1,29 @@
+import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
+import {Platform} from "../models/platform";
+
+type AppleSubscription = {
+    receipt: string
+    originalTransactionId: string
+}
+
+export type AppleLinkPayload = {
+    platform: Platform.DailyEdition | Platform.Ios | Platform.IosPuzzles | Platform.IosEdition,
+    subscriptions: AppleSubscription[]
+}
+
+function deduplicate<T, U>(list: T[], selector: (item: T) => U): T[] {
+    return list.reduce<T[]>(
+        (agg, item) =>
+            agg.some((x) => selector(x) === selector(item)) ?
+                agg : agg.concat([item]), 
+        []
+    )
+}
+
+export function parseAppleLinkPayload(request: APIGatewayProxyEvent): AppleLinkPayload {
+    const parsed = JSON.parse(request.body ?? "") as AppleLinkPayload;
+    return {
+        ...parsed,
+        subscriptions: deduplicate(parsed.subscriptions, x => x.originalTransactionId)
+    }
+}
--- a/typescript/src/link/link.ts
+++ b/typescript/src/link/link.ts
@@ -71,23 +71,23 @@ interface IdentityResponse {
 }
 
 function consentPayload(): any {
-    [
-        {
-           "id" : "guardian_weekly_newsletter",
-           "consented" : false
-         },
-         {
-           "id" : "similar_guardian_products",
-           "consented" : true
-         },
-         {
-           "id" : "subscriber_preview",
-           "consented" : true
-         },
-         {
-           "id" : "supporter_newsletter",
-           "consented" : true
-         }
+   [
+    {
+       "id" : "your_support_onboarding",
+       "consented" : true
+     },
+     {
+       "id" : "similar_guardian_products",
+       "consented" : true
+     },
+     {
+       "id" : "digital_subscriber_preview",
+       "consented" : true
+     },
+     {
+       "id" : "supporter_newsletter",
+       "consented" : true
+     }
    ]
 }
 
--- a/typescript/src/link/link.ts
+++ b/typescript/src/link/link.ts
@@ -140,6 +140,16 @@ export async function postConsent(identityId: string, identityToken: string): Pr
     This will clearly identify the code that needs to be modified, cleaned up later.
 */
 
+const soft_opt_in_v1_active: boolean = false;
+
+/*
+    Date: March 2023, 6th
+    Author: Pascal
+
+    Introduced the `soft_opt_in_v1_5_active` variable above to guard against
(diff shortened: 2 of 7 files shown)


typescript/src/link/link.ts
```diff
@@ -59,6 +61,89 @@ async function persistUserSubscriptionLinks(userSubscriptions: UserSubscription[
...
+interface UserId {
+    id: string
+}
+
+interface IdentityResponse {
+    status: string,
+    user: UserId
+}
+
+function consentPayload(): any {
```
jonflynng, line 73: Can you replace this JSON with the following:

`[
     {
        "id" : "your_support_onboarding",
        "consented" : true
      },
      {
        "id" : "similar_guardian_products",
        "consented" : true
      },
      {
        "id" : "digital_subscriber_preview",
        "consented" : true
      },
      {
        "id" : "supporter_newsletter",
        "consented" : true
      }
    ]
    `
    
  It might need to be changed once more, we'll receive full confirmation soon.

shtukas: Done, and no worries if it needs to change again later :)

KentGrm (approved):

typescript/src/link/apple-utils.ts
```diff
@@ -0,0 +1,29 @@
+import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
+import {Platform} from "../models/platform";
+
+type AppleSubscription = {
+    receipt: string
+    originalTransactionId: string
+}
+
+export type AppleLinkPayload = {
+    platform: Platform.DailyEdition | Platform.Ios | Platform.IosPuzzles | Platform.IosEdition,
```
graham228221, line 10: So does this fire for ALL mobile subscriptions - the main app but also puzzles and editions?

If so, I'm not sure if we've considered this!

shtukas:
I just made this one https://github.com/guardian/mobile-purchases/pull/894 to simply 889.

shtukas:
The way the `typescript/src/link/apple.ts` file was written, the jest parser would eventually find the `import fetch from 'node-fetch';` and complain that it's not real JavaScript. With this refactoring we allow for the test of function `parseAppleLinkPayload` without importing the entire universe. (It is actually possible, if not even desirable to continue that refactoring and push non handler functions into their own file, but this is not a necessity for the moment).

typescript/src/link/apple-utils.ts
shtukas: Just the main app for v1 (not puzzles etc.)

typescript/src/link/link.ts
```diff
@@ -59,6 +61,85 @@ async function persistUserSubscriptionLinks(userSubscriptions: UserSubscription[
...
+     {
+       "id" : "supporter_newsletter",
+       "consented" : true
+     }
+   ]
+}
+
+export async function postConsent(identityId: string, identityToken: string): Promise<boolean> {
+    var url = `http://idapi.code.dev-theguardian.com/${identityId}/consents`
+    if (Stage === "PROD") {
```
SarahJessica, line 92: I would prefer to see an enum rather than a string but I understand this is convention.

SarahJessica (approved):

typescript/src/link/link.ts
```diff
@@ -80,9 +161,19 @@ export async function parseAndStoreLink<A, B>(
...
                 case "success": {
-                    const insertCount = await persistUserSubscriptionLinks(toUserSubscription((resolution.userId as string), payload));
+                    const userId = resolution.userId as string;
+
+                    const insertCount = await persistUserSubscriptionLinks(toUserSubscription(userId, payload));
                     const sqsCount = await enqueueUnstoredPurchaseToken(toSqsPayload(payload));
                     console.log(`Put ${insertCount} links in the DB, and sent ${sqsCount} subscription refs to SQS`);
+
+                    // ---------------------------------------
+                    // Soft Opt-In version 1
```
paulbrown1982, line 171: Is there a way to put this behind a feature switch so that we can roll this PR out now and release this SOI v1 later?

shtukas: Actually let me do that now. In either case a PR will need to be made for go live since it's backend code.

paulbrown1982 (approved):
👍  LGTM

typescript/src/link/link.ts
```diff
@@ -80,9 +171,18 @@ export async function parseAndStoreLink<A, B>(
...
                 }
                 case "success": {
-                    const insertCount = await persistUserSubscriptionLinks(toUserSubscription((resolution.userId as string), payload));
+                    const userId = resolution.userId as string;
+
+                    const insertCount = await persistUserSubscriptionLinks(toUserSubscription(userId, payload));
                     const sqsCount = await enqueueUnstoredPurchaseToken(toSqsPayload(payload));
                     console.log(`Put ${insertCount} links in the DB, and sent ${sqsCount} subscription refs to SQS`);
+
+                    if (soft_opt_in_v1_active) {
```
paulbrown1982, line 180: @shtukas I only just realised this morning, that for the time being there should be an additional check here which tests that the acquisition date of the IAP matches the current date. 

Eventually, the else condition of this new test (where acquisition of the IAP was yesterday or before) may need to trigger some Braze-based comms to inform the user that a latent account linkage has occurred and maybe that a new SOI situation has been set up.

KentGrm: Might be better to have a historical window? For cases such as when a user buys at midnight or near that time or there is a 4 hour delay because of some kind of technical issue blocking the flow of events.

PR #899, opened 2023-03-08
Soft Opt In version 1 (update 1)

shtukas:
## What does this change?

(... to be written ...)

### DIFF
--- a/typescript/src/link/link.ts
+++ b/typescript/src/link/link.ts
@@ -87,7 +87,7 @@ function consentPayload(): any {
    ]
 }
 
-export async function postConsent(identityId: string, identityToken: string): Promise<boolean> {
+async function postSoftOptInConsent(identityId: string, identityToken: string): Promise<boolean> {
     var url = `http://idapi.code.dev-theguardian.com/${identityId}/consents`
     if (Stage === "PROD") {
         url = `https://idapi.theguardian.com/user/${identityId}/consents`
@@ -117,28 +117,15 @@ export async function postConsent(identityId: string, identityToken: string): Pr
     }
 }
 
-/*
+async function shouldPostSoftOptInConsent(subscriptionId: string): Promise<boolean> {
+    // Write code....
+    return Promise.resolve(false);
+}
 
-    Date: March 2023, 1st
-    Author: Pascal
-
-    At the time these lines are written the Engine team and friends from Retention are
-    working on the Soft Opt-In project, and more exactly what is known as "version/stage 1"
-    of that project.
-
-    The effect for mobile-purchases, is that we are asked to send a payload to the identity API
-    for each acquisition notification from the users mobile apps hitting the endpoints
-        /google/linkToSubscriptions
-        /google/linkToSubscriptions
-
-    This change is... temporary and a stepping stone to what is going to be version 1.5
-    and then later on version 2.
-
-    Some code have been added to this file to support this (temporary) feature. The entry
-    point is what is labelled "Soft Opt-In version 1" in parseAndStoreLink.
-
-    This will clearly identify the code that needs to be modified, cleaned up later.
-*/
+async function updateDynamoTable(subcriptionId: string): Promise<null> {
+    // Write code....
+    return Promise.resolve(null);
+}
 
 const soft_opt_in_v1_active: boolean = false;
 
@@ -146,8 +133,8 @@ const soft_opt_in_v1_active: boolean = false;
     Date: March 2023, 6th
     Author: Pascal
 
-    Introduced the `soft_opt_in_v1_5_active` variable above to guard against
-    the effect as this goes to the main branch. When we go live, just remove the check.
+    Introduced the `soft_opt_in_v1_active` variable above to guard against
+    the effect of the Soft Opt In patch code until it's time to go live.
 */
 
 export async function parseAndStoreLink<A, B>(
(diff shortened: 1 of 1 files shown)


dynamo.cloudformation.yaml
```diff
@@ -160,6 +160,40 @@ Resources:
         - Key: App
           Value: !Ref App
 
+  SoftOptInsLoggingTable:
```
jonflynng, line 163: When deploying to riffraff, this change did not create a new table in AWS. to investigate @shtukas

shtukas:
For testing, can acquire an Okta token from here: https://github.com/guardian/identity/tree/main/docs/postman

shtukas:
Screenshot 2023-03-16 at 11 39 29

shtukas:
Testing data:

shtukas:
Screenshot 2023-03-16 at 14 58 43

shtukas:
Screenshot 2023-03-16 at 14 59 10

shtukas:
Screenshot 2023-03-16 at 14 59 14

shtukas:
Screenshot 2023-03-16 at 14 59 19

shtukas:
Screenshot 2023-03-16 at 15 31 22

shtukas:
Screenshot 2023-03-16 at 15 01 03

shtukas:
Screenshot 2023-03-16 at 15 01 03

typescript/src/link/link.ts
```diff
@@ -87,67 +89,84 @@ function consentPayload(): any {
...
+    } catch (error) {
+        console.warn(error);
+        console.warn(`dynamo write failed for record: ${record}`);
+        await putMetric("failed_consents_updates", 1)
+    }
+}
 
-    This will clearly identify the code that needs to be modified, cleaned up later.
-*/
+async function getIdenityApiKey(): Promise<string> {
```
graham228221, line 158:
```suggestion
async function getIdentityApiKey(): Promise<string> {
```

typescript/src/link/link.ts
```diff
@@ -175,14 +194,88 @@ export async function parseAndStoreLink<A, B>(
...
+                            because it doesn't carry the right scopes.
+
+                        */
+
+                        if (softOptInQueryParameterIsPresent(httpRequest)) {
+                            console.log(`softOptInQueryParameterIsPresent`);
+                            const userAuthenticationToken = getAuthToken(httpRequest.headers) as string;
+                            const subscriptionsFromHttpPayload = toUserSubscription(userId, payload);
+
+                            const identityApiKey = await getIdenityApiKey();
```
graham228221, line 265:
```suggestion
                            const identityApiKey = await getIdentityApiKey();
```

dynamo.cloudformation.yaml
```diff
@@ -160,6 +160,40 @@ Resources:
...
+    Type: AWS::DynamoDB::Table
+    Properties:
+      TableName: !Sub ${App}-${Stage}-soft-opt-ins-logging
```
graham228221, line 166: Should this be `-soft-opt-ins-logging-v2`? If not, why is this table needed? I think only the `v2` table is [referenced](https://github.com/guardian/mobile-purchases/pull/899/files#diff-e0c1007af25cb5be3ebada448c237f7941be498badbea7856d26753036da9542R26).

shtukas: @jonflynng

andrewHEguardian: I can try answer this - basically we deployed to CODE with `soft-opt-ins-logging` but with the wrong AttributeDefinitions. When redeploying with the updated cloudformation the policy did not allow us to delete and recreate the table, so as a workaround Jon created `soft-opt-ins-logging-v2`. So not ideal

graham228221: Ah ok, sounds like something to come back to later on.

typescript/src/link/link.ts
```diff
@@ -175,14 +194,88 @@ export async function parseAndStoreLink<A, B>(
...
+
+                            We are using and have created mp-soft-opt-in-identity-api-key just for this.
+                            We could not use the userAuthenticationToken to post the consent object to Identity
+                            because it doesn't carry the right scopes.
+
+                        */
+
+                        if (softOptInQueryParameterIsPresent(httpRequest)) {
+                            console.log(`softOptInQueryParameterIsPresent`);
+                            const userAuthenticationToken = getAuthToken(httpRequest.headers) as string;
```
graham228221, line 262:
```suggestion
                            // const userAuthenticationToken = getAuthToken(httpRequest.headers) as string;
```

Maybe comment this out if it's not being used (or remove if it's not going to be bought back later)

typescript/src/link/link.ts
```diff
@@ -87,67 +89,84 @@ function consentPayload(): any {
...
         body: JSON.stringify(consentPayload()),
         headers: {
-            Authorization: `Bearer ${identityToken}`,
+            Authorization: `Bearer ${identityApiKey}`,
             'Content-type': 'application/json',
         }
     }
     try {
+        console.log(`url ${url}`);
+        console.log(`identityApiKey ${identityApiKey}`);
```
graham228221, line 107: I feel like we should be keeping API secret keys out of the logs, what do we think?

KentGrm (approved):

typescript/src/link/link.ts
shtukas: Oh yes. Good idea. Will do this now.

jonflynng:
How are we fetching the idapi key? We could store it in secrets manager and add it to the cloudformation so it get's loaded into the environment variable we're querying on?

jonflynng:
Is it fetched from S3 and stored in memory?

PR #941, opened 2023-03-30
Clean up Cloudformation

jonflynng:
We have two tables in the `cloudformation.yaml` due to the inability to change keys in an already created dynamo table. This PR removes both tables from this file. Instead, I will upload the `dynamo.cloudformation.yaml` manually to the `mobile-purchases-dynamo` stack to keep consistency with all the other dynamoDB tables. 

- [ ] Tested in DEV


shtukas (approved):
Assuming successful tests in CODE

PR #942, opened 2023-03-30
move identity api user consent url to parameter store

shtukas:
We have created two entries in the parameter store for the primary url for identity user consent.

```
key  : /mobile-purchases/CODE/mobile/mp-soft-opt-in-identity-user-consent-domain-url
value: https://idapi.code.dev-theguardian.com

key  : /mobile-purchases/PROD/mobile/mp-soft-opt-in-identity-user-consent-domain-url
value: https://idapi.theguardian.com
```

We can then refactor the code with a call to `getConfigValue`.


shtukas:
Screenshot 2023-03-30 at 15 56 36

PR #952, opened 2023-04-06
Soft opt ins for IAPS v1.5+: Cancellations

jonflynng:
**This PR has been forked off [this one](https://github.com/guardian/mobile-purchases/pull/900/files). I'll make sure I resolve any comments in that one**

## What does this change?

This PR adds functionality to send an SQS message cross-account to the `soft-optin-consent-setter-queue` in the `membership` account when a subscription is cancelled, ensuring that the user's soft opt-ins are disabled accordingly. We use an existing DynamoDB stream from the `subscriptions` table to react to a non-null cancellationTimestamp after a record is modified.

**Key changes:**

 - Changed the DynamoDB stream's current specification from `KEYS_ONLY` to `OLD_AND_NEW_IMAGE`.
- Added a new handler function for processing cancelled subscriptions.
- Because we are changing a currently active dynamoDB stream, we need to follow an iterative deployment process to production, as described below:

    1.  Update the dependent stack(s) (e.g., 'mobile-purchases-${Stage}') to use a temporary value for the DynamoDB stream ARN or remove the dependency temporarily. [PR for this step here](https://github.com/guardian/mobile-purchases/pull/953).
    2.   Update the original stack with the new StreamSpecification for the DynamoDB table. [PR for this step here](https://github.com/guardian/mobile-purchases/pull/954).
    3.   Update the dependent stack(s) again to use the correct DynamoDB stream ARN from the original stack.
    
**Important:** This PR is step 3 in this process, the other two PR's must be merged before!

### Before deploying

 We need to manually upload the ZIP file containing the compiled code for this lambda so Cloudformation does not throw an error. If we look at the riffraff.yaml, it updates the Cloudformation first before the `aws-lambda` deployment, therefore it would result in a `404` error when attempting to grab the code from the `S3` bucket. I can amend this as a piece of health work in the future, perhaps when migrating the build from teamcity to github actions.

This PR is dependent on IAM roles in these PR's:

https://github.com/guardian/membership-workflow/pull/399
https://github.com/guardian/support-service-lambdas/pull/1920

- [x] Create folder in S3 for new lambda and upload ZIP
- [x] Merge PR with IAM role in support-service-lambdas
- [x] Merge PR with IAM role in membership-workflow
- [x] Merge this [PR](https://github.com/guardian/mobile-purchases/pull/953) for the first step of the deployment process
- [x] Merge this [PR](https://github.com/guardian/mobile-purchases/pull/954) for the second step of the deployment process

PR #953, opened 2023-04-06
Deploy new stream spec to prod 1

jonflynng:
Part of the [soft opt ins for IAP's work](https://github.com/guardian/mobile-purchases/pull/952).

As part of the Soft Opt-Ins for IAP's cancellation work, we have to alter a currently active dynamoDB stream. We need to follow an iterative deployment process to production to alter a currently active stream, detailed below:

  1. Update the dependent stack(s) (e.g., 'mobile-purchases-${Stage}') to use a temporary value for the DynamoDB stream ARN or remove the dependency temporarily.
  2. Update the dynamodb stack with the new StreamSpecification for the DynamoDB table.
  3. Update the dependent stack(s) again to use the correct DynamoDB stream ARN from the original stack.
    
This PR is the first step in this process. It adds compatibility for both the old and new stream specs in the `deleteLink` file.


jonflynng:
[Soft Opt-Ins for IAP's V1.5+: Cancellations](https://trello.com/c/wfpMxmNU/4026-soft-opt-ins-for-iaps-v15-cancellations)

andrewHEguardian (approved):
👍

PR #973, opened 2023-04-20
Soft Opt-Ins cancellations tweaks and fixes

jonflynng:
This PR contains added permissions and a fix in the logic filtering the records we want to process.

### DIFF
--- a/typescript/src/soft-opt-ins/cancelSubscription.ts
+++ b/typescript/src/soft-opt-ins/cancelSubscription.ts
@@ -35,8 +35,6 @@ export async function handler(
 	event: DynamoDBStreamEvent,
 ) {
 	try {
-		event.Records.filter(x => console.log(JSON.stringify(x)));
-
 		const cancellationEvents = getCancellationRecords(event);
 
 		console.log(`${cancellationEvents.length} records to process`)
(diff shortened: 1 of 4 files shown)


typescript/src/soft-opt-ins/cancelSubscription.ts
```diff
@@ -21,29 +21,42 @@ async function getUserSubscription(subscriptionId: string): Promise<UserSubscrip
 }
 
 export function getCancellationRecords(event: DynamoDBStreamEvent) {
-	return event.Records.filter(dynamoEvent => dynamoEvent.eventName === "MODIFY" &&
```
jonflynng, line 24: This change asserts that only modifications with a new cancellation timestamp are included, otherwise, this would re-trigger on any new modification where a `cancellationTimestamp` is already on the record.

typescript/src/soft-opt-ins/cancelSubscription.ts
```diff
@@ -21,29 +21,42 @@ async function getUserSubscription(subscriptionId: string): Promise<UserSubscrip
...
+			(!oldImage?.cancellationTimestamp || oldImage?.cancellationTimestamp?.S === undefined) &&
+			newImage?.cancellationTimestamp && newImage?.cancellationTimestamp?.S !== undefined;
+	});
 }
 
 export async function handler(
 	event: DynamoDBStreamEvent,
 ) {
 	try {
+		event.Records.filter(x => console.log(JSON.stringify(x)));
```
andrewHEguardian, line 38: why is this a filter?

jonflynng: ah , ill remove, that was for debugging purposes

andrewHEguardian (approved):

PR #977, opened 2023-04-20
Soft opt ins additional functionality

jonflynng:
We recently merged the V1.5+ functionality for processing acquisitions, post acquisition sign-ins, and cancellations into production (behind a feature flag). This uses DynamoDB streams to forward messages to the soft-opt-in-consent-setter lambda. However, we still need to consider two more scenarios: 

 1. Subscriptions that expired due to payment failure or other reasons (no explicit cancellation) (Processed in the `deleteLink` file)
 2. Uncancellations (Processed in the `cancelSubscription` file)

### DIFF
--- a/typescript/src/link/deleteLink.ts
+++ b/typescript/src/link/deleteLink.ts
@@ -1,11 +1,18 @@
 import 'source-map-support/register'
 import {DynamoDBStreamEvent} from "aws-lambda";
-import {dynamoMapper} from "../utils/aws";
+import {dynamoMapper, sendToSqsSoftOptIns} from "../utils/aws";
 import {ReadUserSubscription} from "../models/userSubscription";
+import {getMembershipAccountId} from "../utils/guIdentityApi";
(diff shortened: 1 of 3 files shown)


typescript/src/link/deleteLink.ts
```diff
@@ -1,11 +1,18 @@
...
+import {Region, Stage} from "../utils/appIdentity";
+
+async function getUserLinks(subscriptionId: string) {
+    const userLinks = await dynamoMapper.query(ReadUserSubscription, {subscriptionId}, {indexName: "subscriptionId-userId"});
+    return userLinks;
+}
 
 async function deleteUserSubscription(subscriptionId: string): Promise<number> {
     let count = 0;
-    const userLinks = await dynamoMapper.query(ReadUserSubscription, {subscriptionId}, {indexName: "subscriptionId-userId"});
```
jonflynng, line 8: just some refactoring so we don't call this query twice

shtukas (approved):

PR #978, opened 2023-04-21
Remove cancellations logic

jonflynng:
After discussion with DRR, we have decided to only disable soft opt-ins when the subscription has expired, not on cancellation/uncancellation. This is so we can send additional comms during the cancellation pending period.


andrewHEguardian (approved):
🗑️ 👍

shtukas (approved):

PR #981, opened 2023-04-21
Expiration improvements and fixes

jonflynng:
**Changes:**

- Added logging of soft opt-in updates to the `SoftOptInsLoggingTable`
- Added unit tests for the `deleteLink` function
- Added appropriate permissions to the lambda and simplified cloudformation by removing the `DeleteUserLinkLambdasRole`.
- Optimised dynamo queries by only calling the `getUserLinks` function once per `subscriptionId`.
- Added alarm for soft opt-ins and error handling

- [x] Tested in DEV

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -910,51 +910,6 @@ Resources:
         - Ref: AlarmTopic
       TreatMissingData: notBreaching
 
-  DeleteUserLinkLambdasRole:
-    Type: AWS::IAM::Role
-    Properties:
-      AssumeRolePolicyDocument:
-        Statement:
-          - Effect: Allow
-            Principal:
-              Service:
-                - lambda.amazonaws.com
-            Action: sts:AssumeRole
-      Path: /
-      Policies:
-        - PolicyName: logs
-          PolicyDocument:
-            Statement:
-              Effect: Allow
-              Action:
-                - logs:CreateLogGroup
-                - logs:CreateLogStream
-                - logs:PutLogEvents
-                - cloudwatch:putMetricData
-              Resource: "*"
-        - PolicyName: dynamoStream
-          PolicyDocument:
-            Statement:
-              Effect: Allow
-              Action:
-                - "dynamodb:GetRecords"
-                - "dynamodb:GetShardIterator"
-                - "dynamodb:DescribeStream"
-                - "dynamodb:ListStreams"
-              Resource:
-                Fn::ImportValue:
-                  !Sub ${App}-${Stage}-subscriptions-stream-arn
-        - PolicyName: dynamo
-          PolicyDocument:
-            Statement:
-              Effect: Allow
-              Action:
-                - "dynamodb:Query"
-                - "dynamodb:DeleteItem"
-              Resource:
-                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions
-                - !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${App}-${Stage}-user-subscriptions/*
-
   AppleRevalidateSubscriptionRole:
     Type: AWS::IAM::Role
     Properties:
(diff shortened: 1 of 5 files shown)


typescript/src/soft-opt-ins/acquisitions.ts
```diff
@@ -29,7 +29,7 @@ async function updateDynamoLoggingTable(subcriptionId: string, identityId: strin
     }
 }
 
-async function handleError(identityId: string, message: string): Promise<never> {
```
jonflynng, line 32: this parameter wasn't being used so have deleted it and have improved error messages below

typescript/src/utils/aws.ts
```diff
@@ -126,7 +126,14 @@ export function sendToSqs(queueUrl: string, event: any, delaySeconds?: number):
         DelaySeconds: delaySeconds
     }).promise()
 }
-export async function sendToSqsSoftOptIns(queueUrl: string, event: any, delaySeconds?: number): Promise<PromiseResult<Sqs.SendMessageResult, AWSError>> {
+
+interface SoftOptInEvent {
```
jonflynng, line 130: Added type safety

cloudformation.yaml
```diff
@@ -1025,6 +979,52 @@ Resources:
         Stage: !Ref Stage
         Stack: !Ref Stack
         App: !Ref App
+      Policies:
```
andrewHEguardian, line 982: just curious why did you move these from a role to policies?

typescript/src/link/deleteLink.ts
```diff
@@ -1,51 +1,69 @@
...
+
+async function updateDynamoLoggingTable(subcriptionId: string, identityId: string) {
+    const timestamp = new Date().getTime();
+    const record = new SoftOptInLog(identityId, subcriptionId, timestamp, "Soft opt-ins processed for expired subscription");
+
+    try {
+        await dynamoMapper.put({item: record});
+        console.log(`Logged soft opt-in setting to Dynamo`);
+    } catch (error) {
+        handleSoftOptInsError(`Dynamo write failed for record: ${record}`);
```
andrewHEguardian, line 17: should we log the error?

jonflynng: yep, have added these in.

cloudformation.yaml
jonflynng: Simplicity in the cloudformation. Rather than having two resources, an IAM role and a lambda I think it's simpler to just have all the permissions under the lambda.

typescript/src/link/deleteLink.ts
```diff
@@ -60,24 +78,31 @@ export async function handler(event: DynamoDBStreamEvent): Promise<any> {
...
     let records = 0;
     let rows = 0;
-    for await (const deletedCount of subscriptionsPromises) {
+
+    const featureFlag = false;
+
+    for (const subscriptionId of subscriptionIds) {
+        const userLinksIterator = await getUserLinks(subscriptionId);
+
+        const userSubscriptions: ReadUserSubscription[] = [];
```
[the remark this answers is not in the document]
andrewHEguardian: I'm a bit confused, how are we only calling the `getUserLinks` function once after this change?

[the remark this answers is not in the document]
jonflynng: Ah, once per `subscriptionId` that is. Before it was being called twice per `subscriptionId`.

typescript/tests/link/deleteLink.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        const result = await handler(event);
+
+        expect(mockDataMapper.query).toHaveBeenCalledTimes(1);
+        expect(mockDataMapper.query).toHaveBeenCalledWith(ReadUserSubscription, { subscriptionId: "1" }, { indexName: "subscriptionId-userId" });
+
+        expect(mockDataMapper.delete).toHaveBeenCalledTimes(1);
+        expect(mockDataMapper.delete).toHaveBeenCalledWith({"subscriptionId": "1", "userId": "123"});
+
+        /* Following is commented out as the feature flag for soft opt-ins is switched off:
```
andrewHEguardian, line 121: Would be nice to have this independent of the feature flag in the lambda, ie to be able to set it on in the test

andrewHEguardian (approved):
Looks good 👍 nice refactor

shtukas (approved):

PR #1017, opened 2023-05-12
Logging fix

jonflynng:
Was logging the amount of subscriptions processed regardless if disabling soft opt-ins was successful or not. Have added a simple variable to track successes.


andrewHEguardian (approved):
👍

PR #1054, opened 2023-06-08
Remove sfcontactid from SQS SOI message

jonflynng:
membership-workflow is breaking because we are including sfcontactid and then it attempts to retrieve the salesforce contact.

I have just tested removing the property and it works


andrewHEguardian (approved):
👍

PR #1061, opened 2023-06-09
Check subscription is valid in the acquisitions lambda

jonflynng:
Many of the subscriptions being inserted into the link table have an `endTimestamp` in the past. I originally thought only active subs were in this link table, hence why they have a TTL property on their expire date. However, this is not the case. This PR ensures only active subscriptions have soft opt-ins set.

### DIFF
--- a/typescript/src/soft-opt-ins/acquisitions.ts
+++ b/typescript/src/soft-opt-ins/acquisitions.ts
@@ -5,7 +5,7 @@ import {Region, Stage} from "../utils/appIdentity";
 import fetch from 'node-fetch';
 import { Response } from 'node-fetch';
 import {getIdentityApiKey, getIdentityUrl, getMembershipAccountId} from "../utils/guIdentityApi";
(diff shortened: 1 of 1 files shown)


typescript/src/soft-opt-ins/acquisitions.ts
```diff
@@ -94,9 +84,23 @@ async function processAcquisition(record: DynamoDBRecord): Promise<void> {
     try {
         subscriptionRecord = await dynamoMapper.get(itemToQuery);
     } catch (error) {
-        console.log("Subscription record not found in the subscriptions table. Assuming the customer has purchased it just now. Error: ", error);
```
jonflynng, line 97: I'm still not sure why some subscriptions are not found in the subscriptions table. It appears the majority yesterday were inactive subscriptions. I will investigate this today.

shtukas (approved):

andrewHEguardian (approved):
👍

PR #1076, opened 2023-06-16
Run lambda every hour instead of every 3 hours

jonflynng:


andrewHEguardian (approved):
I had to look this syntax up but looks correct to me

PR #1135, opened 2023-08-08
logging

shtukas:


shtukas:
Captured notifications and in particular saw that there seems to be only 4 types: 
- DID_RENEW
- DID_CHANGE_RENEWAL_STATUS
- INITIAL_BUY
- DID_FAIL_TO_RENEW

PR #1160, opened 2023-08-30
/apple/fetchOfferDetails

shtukas:
This PR adds a new end-point to the mobile-purchases family to perform server side signatures for iOS promotional offers (part of the 2023 Apps price rise offering), according to the specifications presented here: https://developer.apple.com/documentation/storekit/in-app_purchase/original_api_for_in-app_purchase/subscriptions_and_offers/generating_a_signature_for_promotional_offers

The two end points are 
CODE: https://mobile-purchases.mobile-aws.code.dev-guardianapis.com/apple/fetchOfferDetails
PROD: https://mobile-purchases.mobile-aws.guardianapis.com/apple/fetchOfferDetails

And an example of use is given by

```
curl -X POST -d '{"username": "hashedusername", "productIdentifier": "product1", "offerIdentifier": "offer1"}' https://mobile-purchases.mobile-aws.code.dev-guardianapis.com/apple/fetchOfferDetails
{"nonce":"8ef8a6f8-2916-4c4a-a7d6-ad97258ee56f","timestamp":1693856897490,"keyIdentifier":"PGRFM5F82T","signature":"MEUCIQCoaq1pzCQW6wsf7ga6iareWJ41dvsRXL+LlbO/JgNM+wIgFLwty/Ycz2nDXD4vhBxPJTddwg+Mo9v8gn4T+z4DPxY="}
```

New (mobile account) SSM Parameters generated as part of this work

```
/mobile-purchases/CODE/mobile/promotional-offers-appBundleId
/mobile-purchases/PROD/mobile/promotional-offers-appBundleId
/mobile-purchases/CODE/mobile/promotional-offers-encryption-private-key
/mobile-purchases/PROD/mobile/promotional-offers-encryption-private-key
/mobile-purchases/CODE/mobile/promotional-offers-keyIdentifier
/mobile-purchases/PROD/mobile/promotional-offers-keyIdentifier
```

PR #1182, opened 2023-09-18
user lowercase for username in promotional message

shtukas:
This updates the change that we made here: https://github.com/guardian/mobile-purchases/pull/1160, for the signature of promotional offers, to enforce a lowercase username. As specified by the documentation: https://developer.apple.com/documentation/storekit/in-app_purchase/original_api_for_in-app_purchase/subscriptions_and_offers/generating_a_signature_for_promotional_offers


paulbrown1982 (approved):
👍

PR #1184, opened 2023-09-18
Update AppleReceiptInfo with promotional_offer_id and offer_code_ref_name

shtukas:
We update type `AppleReceiptInfo` with two new fields observed in the wild: `promotional_offer_id` and `offer_code_ref_name`.


paulbrown1982 (approved):
👍 thank you for doing this so quickly.

PR #1197, opened 2023-09-28
Additional fields to support subscription IAP changes

paulbrown1982:
## What does this change?
Adding 2 new fields to the `pending_renewal_info` object, which ultimately emits to the `ophan-raw-mobile-subscription-historical-apple` S3 bucket:

- price_consent_status
- price_increase_status

See their definitions in the Apple documentation: https://developer.apple.com/documentation/appstorereceipts/responsebody/pending_renewal_info

## How to test
Deploy today. Check that the S3 bucket tomorrow contains these fields. Ideally all JSON entries should now have the new `price_increase_status` field with "0" value.

The Data Platform `raw.mobile_subscription_historical_apple` is able to inherit this data

The record could become unparsable via `raw.mobile_subscription_historical_apple` with crossed-over fields, although surely no JSON parser expects the fields to be in a certain order?!

## Images

Screenshot 2023-09-28 at 16 01 50

## Accessibility

### DIFF
--- a/typescript/src/services/appleValidateReceipts.ts
+++ b/typescript/src/services/appleValidateReceipts.ts
@@ -14,7 +14,9 @@ export interface PendingRenewalInfo {
     grace_period_expires_date_ms?: string,
     is_in_billing_retry_period?: "0" | "1"
     original_transaction_id: string
-    product_id: string
+    product_id: string,
+    price_consent_status: string,
+    price_increase_status: string
 }
 
 export interface AppleValidatedReceiptServerInfo {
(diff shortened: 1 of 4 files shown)


typescript/src/services/appleValidateReceipts.ts
```diff
@@ -14,7 +14,9 @@ export interface PendingRenewalInfo {
     grace_period_expires_date_ms?: string,
     is_in_billing_retry_period?: "0" | "1"
     original_transaction_id: string
-    product_id: string
+    product_id: string,
+    price_consent_status: string,
```
paulbrown1982, line 18: 🤦 thank you!

shtukas (approved):

PR #1204, opened 2023-10-04
Set PendingRenewalInfo.price_increase_status as optional

shtukas:
When we did this PR ( https://github.com/guardian/mobile-purchases/pull/1184 ), to update the `AppleReceiptInfo` type with two additional field, we set up `offer_code_ref_name` as mandatory. This was a mistake because they are optionals. (The Apple documentation seems to take the convention that unless specifically declared as mandatory any field is optional). 

Here we update `AppleReceiptInfo` to set `offer_code_ref_name` as optional.


paulbrown1982 (approved):
👍

PR #1311, opened 2024-01-18
Return a stubbed success response for Google subscription status checks

tomwadeson:
## What does this change?

Our Android subscribers are currently in danger of having their experience in the app negatively affected due to underlying issues stemming from rate limiting in the associated Google APIs (`androidpublisher`). 

We are _temporarily_ mitigating these effects by stubbing the API call and uniformly returning success responses.

### DIFF
--- a/typescript/src/subscription-status/googleSubStatus.ts
+++ b/typescript/src/subscription-status/googleSubStatus.ts
@@ -41,6 +41,8 @@ export async function handler(request: APIGatewayProxyEvent): Promise<APIGateway
         "subscriptionExpiryDate": subscriptionExpiryDate
     }
 
+    console.log("Returning a stub 200 response")
+
     // We are returning a stubbed response as a temporary workaround to a rate limit issue that
     // is currently causing a production incident for our Android subscribers.
     return { statusCode: 200, body: JSON.stringify(responseBody) }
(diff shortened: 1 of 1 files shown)


typescript/src/subscription-status/googleSubStatus.ts
```diff
@@ -33,40 +33,15 @@ function googlePackageName(headers: HttpRequestHeaders): string {
 
 export async function handler(request: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
 
-    const purchaseToken = getPurchaseToken(request.headers);
-    const subscriptionId = getSubscriptionId(request.pathParameters);
-    if (purchaseToken && subscriptionId) {
-        const packageName = googlePackageName(request.headers);
-        console.log(`Searching for valid ${subscriptionId} subscription for Android app with package name: ${packageName}`);
+    const today = new Date()
+    const subscriptionExpiryDate = new Date(today.setMonth(today.getMonth() + 1))
```
tomwadeson, line 37: Confirmed with Mohammad that the app only considers the HTTP status and the `"subscriptionHasLapsed"` field, so the value for `"subscriptionExpiryDate"` should not be relevant.

shtukas (approved):

shtukas:
Happy. And I understand that it's temporary. Good work.

PR #1343, opened 2024-02-06
feat: turn off point in time recovery for subscription events

SarahJessica:
## What does this change?

Turns off point in time recovery for a table where we believe it is not required. 

Cost saving on a very large Dynamo table. 

This table is also backed up in 14 nightly snapshots.


MaelGNM (approved):
👍

shtukas (approved):

PR #1358, opened 2024-02-16
Test Google/Android billing upgrades with real subscriptions in production

tomwadeson:
## What does this change?

This change allows us to test the upgraded versions of the Google/Android billing APIs in a real-world environment outside of the critical path of live user journeys.


michaelbjacobson (approved):
All looks good to me! 👍🏼

MaelGNM (approved):
LGTM!

PR #1376, opened 2024-02-29
Update Node to v20.11.1 (Iron)

michaelbjacobson:
Node v14 is now EOL, and AWS will be removing all support for it soon. This PR upgrades Node to v20.11.1, which is the official recommended version for LTS at the time of writing.


michaelbjacobson:
> Cool! But take my +1 with a grain of salt. Not much experience with TS, so unaware of wider impact of this change (with a bump that large in Scala, I'd have expected to have to bump some dependencies too!)

I assumed the same, but there's a fairly comprehensive suite of TypeScript unit tests, and the update hasn't caused any failures, or even warnings!

PR #1388, opened 2024-03-11
Mj/pipe existing cfn through cdk

michaelbjacobson:
This PR uploads the existing (YAML) CFN template via GuCDK. This is the next step in gradually moving over to CDK for the entire project.


AndreaDiotallevi:
Nice, I like the incremental PR approach

AndreaDiotallevi (approved):

tjmw (approved):
Nice, this looks good to me!

Only because Tom Wadeson has spent more time in this repo recently than the rest of us, but I wonder if it's worth waiting for his 👍  when he's back tomorrow.

> This is the next step in gradually moving over to CDK for the entire project.

Yeah that'd be great! I reckon that's probably work for down the line after the Feast specific stuff has been completed.

PR #1396, opened 2024-03-13
Publish Feast/Apple sub notifications to SQS

tomwadeson:
## What does this change?

This PR extends the Feast/Apple webhook handler to publish subscription events to a new SQS queue.

## How to test

Post a dummy webhook payload to the endpoint (use the example from the unit test) and observe that the associated receipt is published to the queue.

Observed in CODE:

Screenshot 2024-03-13 at 14 15 38

## Accessibility

### DIFF
--- a/typescript/tests/feast/pubsub/pubsub.test.ts
+++ b/typescript/tests/feast/pubsub/pubsub.test.ts
@@ -32,7 +32,7 @@ describe("The Feast Apple pubsub", () => {
                     is_trial_period: "true",
                     item_id: "TEST",
                     original_transaction_id: "TEST",
-                    product_id: "uk.co.guardian.gla.12months.2018Dec.withFreeTrial",
+                    product_id: "some.product.id",
                     quantity: "1",
                     transaction_id: "TEST",
                     unique_identifier: "TEST",


typescript/tests/feast/pubsub/pubsub.test.ts
```diff
@@ -0,0 +1,89 @@
...
+                environment: "Sandbox",
+                latest_receipt: "TEST",
+                latest_receipt_info: [{
+                    app_item_id: "TEST",
+                    bvrs: "TEST",
+                    is_in_intro_offer_period: "false",
+                    is_trial_period: "true",
+                    item_id: "TEST",
+                    original_transaction_id: "TEST",
+                    product_id: "uk.co.guardian.gla.12months.2018Dec.withFreeTrial",
```
MaelGNM, line 35: This looks like a real product ID, do we care about its value for the test? Should it be replaced with something more generic?

typescript/src/feast/pubsub/pubsub.ts
```diff
@@ -1,6 +1,41 @@
...
+import Sqs from 'aws-sdk/clients/sqs';
+import { sendToSqs } from "../../utils/aws";
+import { AWSError } from "aws-sdk";
+import { PromiseResult } from "aws-sdk/lib/request";
+import { HTTPResponses } from "../../models/apiGatewayHttp";
 
 export async function handler(request: APIGatewayProxyEvent): Promise<APIGatewayProxyResult>  {
     console.log(`[34ef7aa3] ${JSON.stringify(request)}`)
-    return { statusCode: 200, body: JSON.stringify(request) }
+    return processEvent()(request)
```
tjmw, line 12: Super nitpicky, but I wonder if a slightly better name for the outer function application would be something like `buildProcessEvent`, and we could even separate out the calls to make the intention really clear:

```suggestion
    const processEvent = buildProcessEvent();
    return processEvent(request);
```

What do you reckon? (I don't feel strongly - feel free to ignore!)

PR #1409, opened 2024-03-19
Update revalidate lambda to ignore feast

tjmw:
## What does this change?

When revalidating receipts from Apple, we'll want to handle Feast separately (it uses a different queue, and the lambda which processes items from the queue behaves differently).

## How to test

I'll run the lambda in CODE.

## Accessibility


typescript/src/models/endTimestampFilter.ts
```diff
@@ -33,8 +43,7 @@ export class Subscription {
 export class EndTimeStampFilterSubscription extends Subscription {
 
     constructor() {
-        super("", "" , false);
+        super("", "" , false, "");
```
tjmw, line 46: I don't know why this calls the parent constructor with empty values, but I'm following the existing pattern for the new field.

PR #1413, opened 2024-03-20
add apple feast pubsub secret to cloudformation

michaelbjacobson:
Adds `FeastApplePubSubSecret` param and sets its value as an environment variable in the Feast pubsub lambda.


tjmw (approved):
Great!

PR #1414, opened 2024-03-20
Feast product name should be `FeastInAppPurchase` in soft-opt-in queue

tjmw:
## What does this change?

The soft-opt-in consents for Feast IAPs are different from other IAPs. In the soft-opt-in-consent-setter lambda the config has been updated to specify the new consents, but that lambda needs to be able to distinguish Feast IAPs from others. So in this PR we start providing a different productName (`FeastInAppPurchase`) for Feast events.

## How to test

I've extended and run the Jest tests. I'll also deploy to CODE and test there.

Update: I've tested in CODE. There's something going on with the cross account role and permissions to add to the queue so the lambda invocation errors, but I logged the message which would be added to the queue and it looks correct for both non-Feast and Feast:

Screenshot 2024-03-20 at 17 11 23

## Accessibility


tjmw:
> And easy to extend when Platform.AndroidFeast comes along.

Yeah I think so! I guess that'll map to `FeastInAppPurchase` too, so hopefully pretty much a one line change.

PR #1415, opened 2024-03-20
Remove unused global secondary index on subscriptions table

tjmw:
## What does this change?

In #1410 a new index was added to replace the one removed here (it's the same, but also includes the platform). Since #1409 we've been using the new index, so now the old index can be removed.

## Accessibility


ecguardian (approved):
👍

PR #1416, opened 2024-03-21
Define and deploy a new Lambda to evolve our Feast subscription cache in response to events from Apple

tomwadeson:
## What does this change?

This change provides a new Lambda, `${App}-feast-apple-update-subscriptions-${Stage}`, that:

1. Subscribes to the SQS queue `${App}-${Stage}-feast-apple-subscriptions-to-fetch`
2. Decodes subscription receipts from message payloads
3. Looks up the associated subscription by hitting an Apple API
4. Upserts the retrieved subscription into our cache, `{$App}-${Stage}-subscriptions`

## How to test

The unit test cover as much of the behaviour as possible without integration with real services, and I'll update this section when I manage to test this in `CODE`.

## Accessibility

### DIFF
--- a/typescript/src/services/productBillingPeriod.ts
+++ b/typescript/src/services/productBillingPeriod.ts
@@ -25,7 +25,8 @@ export const PRODUCT_BILLING_PERIOD: {[productId: string]: string} = {
     "uk.co.guardian.subscription.8": "P1M",
     "uk.co.guardian.subscription.9": "P6M",
     "uk.co.guardian.Feast.yearly": "P1Y",
+    // "guardain" is an unfortunate typo that exists upstream and must be mirrored here
     "uk.co.guardain.Feast.yearly.discounted": "P1Y",
     "uk.co.guardian.Feast.monthly": "P1M",
-    "uk.co.guardain.Feast.monthly.discounted": "P1M",
+    "uk.co.guardian.Feast.monthly.discounted": "P1M",
 };


typescript/src/models/app.ts
```diff
@@ -0,0 +1,4 @@
+export enum App {
+    Live = "live",
+    Feast = "feast"
+}
```
tjmw, line 4: Would it be more accurate to encode this as `Feast` and `Default` or similar? I think Editions and Puzzles purchases will be looked up using the same method and route as the live app (though I could be wrong!) Or could also model the other apps in here and have the mapping from app to password return a different value for feast, and the default for other apps.

typescript/src/services/appleValidateReceipts.ts
```diff
@@ -225,8 +235,8 @@ async function retryInSandboxIfNecessary(parsedResponse: AppleValidationServerRe
     }
 }
 
-export function validateReceipt(receipt: string, options: ValidationOptions): Promise<AppleValidationResponse[]> {
-    return callValidateReceipt(receipt)
+export function validateReceipt(receipt: string, options: ValidationOptions, app: App = App.Live): Promise<AppleValidationResponse[]> {
+    return callValidateReceipt(receipt, false, app)
```
tjmw, line 239: If you wanted to keep using the default value for `forceSandbox` you could change the order of args to:

```
callValidateReceipt(receipt: string, app: App = App.Live, forceSandbox: boolean = false)
```

Which would allow calling like this:

```
callValidateReceipt(receipt, app);
```

No biggy either way though! I do wonder if we always provide app here, if we should make it a required param?

typescript/tests/feast/update-subs/updatesubs.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        const result =
+            await handler(event)
+
+        expect(mockStoreSubscriptionInDynamo.mock.calls.length).toEqual(3)
+
+        const storedSubscriptionIds =
+            mockStoreSubscriptionInDynamo.mock.calls.map(call => call[0].subscriptionId)
+
+        // The receipts `"TEST_RECEIPT_1"` & `"TEST_RECEIPT_2"` together reference the subscriptions with IDs
```
tjmw, line 23: Would it be worth moving this comment up to line 17? I was confused for a sec what the relevance of the 3 was on line 18, but this comment explains it!

tjmw (approved):
Nice work, this looks really good. I left a few nitpicky comments but no blockers (could easily be ignored, or put off until a future PR!)

typescript/src/services/productBillingPeriod.ts
```diff
@@ -24,4 +24,8 @@ export const PRODUCT_BILLING_PERIOD: {[productId: string]: string} = {
     "uk.co.guardian.subscription.7": "P1M",
     "uk.co.guardian.subscription.8": "P1M",
     "uk.co.guardian.subscription.9": "P6M",
+    "uk.co.guardian.Feast.yearly": "P1Y",
+    "uk.co.guardain.Feast.yearly.discounted": "P1Y",
+    "uk.co.guardian.Feast.monthly": "P1M",
+    "uk.co.guardain.Feast.monthly.discounted": "P1M",
```
tjmw, line 30:
```suggestion
    "uk.co.guardian.Feast.monthly.discounted": "P1M",
```

PR #1422, opened 2024-03-25
Exchange external for internal IDs

tomwadeson:
## What does this change?

This PR:

1. Retrieves an `appAccountToken` from a subscription when looking up the receipt with Apple
2. Exchanges this `appAccountToken` (a Braze UUID) for an identity ID via the Braze API
3. Inserts a record into the `user-subscriptions` table to relate the subscription to an identity user

## How to test

The unit tests exercise the business logic, while the integration has been tested manually in CODE.

## Accessibility

### DIFF
--- a/typescript/src/services/braze.ts
+++ b/typescript/src/services/braze.ts
@@ -27,9 +27,9 @@ export async function getIdentityIdFromBraze(externalId: string): Promise<string
                 if (identityId) {
                     return identityId
                 }
-                throw new Error(`Unable to exchange the Braze ID '${externalId}' for an Identity ID`)
+                throw new Error(`Response from Braze for Braze ID '${externalId}' did not contain an identity_id`)
             } else {
-                throw new Error("Unable to fetch user from Braze")
+                throw new Error("Received a non-ok response from Braze attempting to fetch user")
             }
         })
 }


typescript/tests/services/appleValidateReceipts.test.ts
```diff
@@ -25,6 +25,7 @@ describe("The apple validation service", () => {
             latestReceipt: "cmVjZWlwdA==",
             latestReceiptInfo: {
                 bundleId: "uk.co.guardian.iphone2",
+                appAccountToken: undefined,
```
tjmw, line 28: Not a big deal at all but this _should_ behave the same as omitting the key completely, I wonder if that's preferable to reduce the changes in the PR?

typescript/src/services/braze.ts
```diff
@@ -0,0 +1,35 @@
...
+        .then(async (response) => {
+            if (response.ok) {
+                const json = await response.json();
+                const identityId = json?.users?.[0]?.custom_attributes?.identity_id
+                if (identityId) {
+                    return identityId
+                }
+                throw new Error(`Unable to exchange the Braze ID '${externalId}' for an Identity ID`)
+            } else {
+                throw new Error("Unable to fetch user from Braze")
```
tjmw, line 32: I wonder if in these two errors we could be even more explicit about what went wrong?

```suggestion
                throw new Error(`Response from Braze for Braze ID '${externalId}' did not contain an identity_id`)
            } else {
                throw new Error("Received a non-ok response from Braze attempting to fetch user")
```

Or something like that? What do you reckon?

typescript/src/feast/update-subs/updatesubs.ts
```diff
@@ -35,6 +63,18 @@ export function buildHandler(
                     await fetchSubscriptionsFromApple(reference)
 
                 await Promise.all(subscriptions.map(storeSubscriptionInDynamo))
+
+                const userSubscriptions =
+                    await Promise.all(subscriptions.map(async s => {
+                        const identityId =
+                            await exchangeExternalIdForIdentityId(s.appAccountToken)
```
tjmw, line 70: Should we be throwing a `ProcessingError` at some point around here with the retry flag set to true to ensure a failure to lookup the user in Braze results in a retry?

typescript/tests/feast/update-subs/updatesubs.test.ts
```diff
@@ -51,19 +98,32 @@ const buildSqsEvent = (receipts: string[]): SQSEvent => {
...
+    subscription("sub-4", "TEST_RECEIPT_3", "app-account-token-3", "identity-id-3"),
 ]
+
+const stubFetchSubscriptionsFromApple =
+    (reference: AppleSubscriptionReference) => 
+        Promise.resolve(subscriptions.filter(s => s.subscription.receipt == reference.receipt).map(s => s.subscription))
+
+const stubExchangeExternalIdForIdentityId = 
+    (externalId: string) =>
+        Promise.resolve(subscriptions.find(s => s.subscription.appAccountToken == externalId)?.identityId!)
```
tjmw, line 122: Aha we were discussing this earlier I think, I've never come across this notation before (the `!`) - it's this: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator. So this is telling TS that we're guaranteeing to return a `string` here, rather than `string | undefined` which is what TS infers without the `!`. I was going to suggest that `string | undefined` sounds better, as it's more accurate that we _might_ fail to exchange, but actually if that's the case we'll throw an error (which isn't encoded at the type level like it would be in Scala etc) so I think this is fine.

typescript/src/feast/update-subs/updatesubs.ts
tjmw: Ah fair enough, I made an assumption that we were handling `ProcessingError` somewhere higher up and changing behaviour based on the `retry` setting.

tjmw (approved):

PR #1424, opened 2024-03-26
add secret validation

michaelbjacobson:
Small change to wrap handler logic inside secret validation clause.


tomwadeson (commented):
Nice one!

I guess the tests are failing as the environment variable isn't set in that context... and maybe it's worth while you're in there just adding a test to exercise the `HTTPResponses.UNAUTHORISED` case?

tomwadeson (approved):

michaelbjacobson:
> Nice one!
> 
> I guess the tests are failing as the environment variable isn't set in that context... and maybe it's worth while you're in there just adding a test to exercise the `HTTPResponses.UNAUTHORISED` case?

Good shout, I've just added that test @tomwadeson 👍🏼

PR #1428, opened 2024-03-27
Remove 2 day window soft-opt-in email check for Feast

tjmw:
## What does this change?

Removes the 2 day window soft-opt-in email check for Feast.

This is preventing Feast IAPs from triggering the SOI email. The condition exists for live to guard against data moving through the system slowly, but the live app works differently (uses the /link endpoint for example).

I've commented the test out because I had a hard time updating the mocks to cover this new scenario. I'll fix these in a subsequent PR (or possibly refactor to move away from mocks).

## Accessibility

### DIFF
--- a/typescript/src/soft-opt-ins/processSubscription.ts
+++ b/typescript/src/soft-opt-ins/processSubscription.ts
@@ -130,7 +130,9 @@ export async function processAcquisition(subscriptionRecord: ReadSubscription, i
         handleError(`Soft opt-in message send failed for subscriptionId: ${subscriptionId}. ${e}`)
     }
 
-    if (subscriptionRecord && isPostAcquisition(subscriptionRecord.startTimestamp)) {
+    const isFeast = subscriptionRecord.platform === Platform.IosFeast;
+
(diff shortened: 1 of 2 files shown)


typescript/src/soft-opt-ins/processSubscription.ts
```diff
@@ -130,7 +130,9 @@ export async function processAcquisition(subscriptionRecord: ReadSubscription, i
         handleError(`Soft opt-in message send failed for subscriptionId: ${subscriptionId}. ${e}`)
     }
 
-    if (subscriptionRecord && isPostAcquisition(subscriptionRecord.startTimestamp)) {
+    const isFeast = subscriptionRecord.platform === Platform.IosFeast;
+
+    if (subscriptionRecord && (isPostAcquisition(subscriptionRecord.startTimestamp) || isFeast)) {
```
rupertbates, line 135: Not immediately obvious what the precedence is here, are we still checking that subscriptionRecord is non null?

rupertbates (approved):
With one question

PR #1463, opened 2024-04-23
Feast subscription update lambda: write the subscription record, even if app account token is missing

tjmw:
## What does this change?

We've identified some cases where a receipt for Feast may not contain an `appAccountToken`, which is how we link to an identity ID. Handling these correctly and ensuring they are linked with an identity ID is something we're looking at separately.  In the meantime, the lack of `appAccountToken` currently causes is to fail to write the subscription record AND the link record. But we shouldn't let it prevent us from writing the subscription. It's useful data, and makes it easier for us to identity the records which don't have a link to an identity ID.

## How to test

I've added an integration style test to verify the new behaviour. I've deployed to CODE, and verified that behaviour is good when there _is_ an app account token, but cannot figure out how to trigger the new code path (since this depends on the response from Apple). I suggest getting this into production and keeping a close eye on it.

## Accessibility

### DIFF
--- a/typescript/src/feast/update-subs/updatesubs.ts
+++ b/typescript/src/feast/update-subs/updatesubs.ts
@@ -13,8 +13,10 @@ import { GracefulProcessingError } from "../../models/GracefulProcessingError";
 type AppAccountToken = {
     appAccountToken: string
 }
-
 type HasAppAccountToken<A> = AppAccountToken & A
+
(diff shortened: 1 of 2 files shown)


MaelGNM (approved):
LGTM!

tjmw:
Now this is out on prod, I can see that some records which were previously failing due to lack of an `appAccountToken` now have a record in the subscription Dynamo table.

PR #1461, opened 2024-04-22
Google Pubsub: Log the notification 5% of the time, when we have errors

tjmw:
## What does this change?

Log the notification payload 5% of the time. There are a lot of errors in the logs for this lambda, but not enough detail to debug.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #1465, opened 2024-04-24
Address Android pubsub errors

tjmw:
## What does this change?

In production we see a large volume of errors due to the payload sent to the pubsub endpoint by Google not being what we expect. We expect to always receive a [subscription notification](https://developer.android.com/google/play/billing/rtdn-reference#voided-purchase) and we (before this PR) cast to a type which matches this.

After adding some logging in #1461 I can see that we often receive a [voided purchase notification](https://developer.android.com/google/play/billing/rtdn-reference#voided-purchase). So far I've not been able to get clarity on what we should be doing with these, so let's change the behaviour to fail gracefully (and log) until we've figured out what to do. Now when we receive a voided purchase notification we'll do nothing and return a 200.

I've introduced [zod](https://zod.dev/) to do more explicit validation. I hope to extend this to other areas of the codebase in future PRs.

## Accessibility

### DIFF
--- a/typescript/src/pubsub/google.ts
+++ b/typescript/src/pubsub/google.ts
@@ -8,6 +8,7 @@ import {Option} from "../utils/option";
 import {fromGooglePackageName} from "../services/appToPlatform";
 import {fetchGoogleSubscription, GOOGLE_PAYMENT_STATE} from "../services/google-play";
 import { z } from "zod";
+import { Ignorable } from './ignorable';
 
 const DeveloperNotificationBaseSchema = z.object({
     version: z.string(),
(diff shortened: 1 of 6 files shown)


typescript/src/pubsub/google.ts
```diff
@@ -7,34 +7,68 @@ import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
...
+            return new Error(`HTTP Payload body parse error: ${parseResult.error}`);
+        }
+
+        const data = parseResult.data;
+        if (isSubscriptionNotification(data)) {
+            return data;
         }
-        return parsedNotification;
+
+        return new Ignorable(`Notification is not a subscription notification. Notification was: ${JSON.stringify(data)}`)
```
tjmw, line 71: If we get this far it means 1) validation of the payload succeeded and 2) it's not a suscription notification, we return an `Ignorable` instance with a message. In the called, this causes a 200 to be returned and no further action to be taken. I originally represented this by returning `undefined` but I think this is more meaningful, and allows us to include a message to log too.

MaelGNM (approved):
Looks good to me!

shtukas (approved):

PR #1474, opened 2024-04-30
Add a lambda to handle feast google purchases

MaelGNM:
## What does this change?
This provides the webhook for the store to send purchase informations to. The lambda doesn't do anything yet.

## How to test
Call the endpoint in CODE and it should return a 200 echoing the request body.
I tested with postman and it worked as intended.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -305,6 +305,17 @@ Resources:
               responses:
                 "200":
                   "description": "200 response"
+          "/feast/google/pubsub":
+            post:
+              x-amazon-apigateway-integration:
+                httpMethod: POST
+                type: aws_proxy
+                uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${FeastGooglePubSubLambda.Arn}/invocations
+              consumes: [application/json]
+              produces: [application/json]
+              responses:
+                "200":
+                  "description": "200 response"
           "/healthcheck":
             get:
               responses:
@@ -598,6 +609,36 @@ Resources:
           Type: Api
           Properties:
             Path: "/apple/fetchOfferDetails"
+            Method: POST
+            RestApiId: !Ref MobilePuchasesApi
+      Tags:
+        Stage: !Ref Stage
+        Stack: !Ref Stack
+        App: !Ref App
+
+  FeastGooglePubSubLambda:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: feast-google-pubsub.handler
+      Runtime: nodejs20.x
+      CodeUri:
+        Bucket: !Ref DeployBucket
+        Key: !Sub ${Stack}/${Stage}/${App}-feast-google-pubsub/feast-google-pubsub.zip
+      FunctionName: !Sub ${App}-feastgooglepubsub-${Stage}
+      Role: !GetAtt MobilePurchasesLambdasRole.Arn
+      Environment:
+        Variables:
+          Stage: !Ref Stage
+          Stack: !Ref Stack
+          App: !Ref App
+      Description: Records Google play store events for Feast app
+      MemorySize: 128
+      Timeout: 29
+      Events:
+        PostApi:
+          Type: Api
+          Properties:
+            Path: "/feast/google/pubsub"
             Method: POST
             RestApiId: !Ref MobilePuchasesApi
       Tags:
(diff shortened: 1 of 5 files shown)


cloudformation.yaml
```diff
@@ -975,6 +1021,35 @@ Resources:
...
+          Value: /feast/google/pubsub
+        - Name: Stage
+          Value: !Ref Stage
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: 1
+      MetricName: 5XXError
+      Namespace: AWS/ApiGateway
+      Period: 300
+      Statistic: Sum
+      Threshold: 2
```
johnduffell, line 1050: if the description is indended as "A Http request.... (failed)" then it should be 1.
Also for "1" error, the alarm comes through up to 4 minutes quicker if it's a 1 minute period
```suggestion
      Period: 60
      Statistic: Sum
      Threshold: 1
```

typescript/src/feast/pubsub/google.ts
```diff
@@ -0,0 +1,6 @@
+import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
+
+export async function handler(request: APIGatewayProxyEvent): Promise<APIGatewayProxyResult>  {
+    console.log(`${JSON.stringify(request)}`)
```
johnduffell, line 4: hardly worth adding now, but it makes it slightly easier to review if there's a comment explaining the body
```suggestion
    // placeholder for new lambda
    console.log(`${JSON.stringify(request)}`)
```

cloudformation.yaml
```diff
@@ -975,6 +1021,35 @@ Resources:
...
+        - Ref: AlarmTopic
+      AlarmName: !Sub mobile-purchases-${Stage}-feast-google-pubsub-check-errors
+      AlarmDescription: A HTTP request to the Feast Google pubsub endpoint resulted in a 5XX error.
+      Dimensions:
+        - Name: ApiName
+          Value: !Sub ${App}-${Stage}
+        - Name: Method
+          Value: POST
+        - Name: Resource
+          Value: /feast/google/pubsub
```
johnduffell, line 1041: alarms are notoriously hard to test - I am not sure if it's worth forcing a 5xx error and making sure a metric does end up in the right place with the right dimensions?

cloudformation.yaml
```diff
@@ -305,6 +309,17 @@ Resources:
...
+            post:
+              x-amazon-apigateway-integration:
+                httpMethod: POST
+                type: aws_proxy
+                uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${FeastGooglePubSubLambda.Arn}/invocations
+              consumes: [application/json]
+              produces: [application/json]
+              responses:
+                "200":
+                  "description": "200 response"
```
johnduffell, line 322:
```suggestion
                  "description": "we successfully processed an incoming ,,,,,something.....?"
```

johnduffell (approved):
looks good, just some small comments (I don't know this code base at all, so I was just revewing it to get an idea of it!)

cloudformation.yaml
MaelGNM: Yes, I'm wondering what we actually want from this alarm. We have the same alarm rule for all the pubsub endpoints, and we'd probable want to keep them the same whichever way we end up going.

typescript/src/feast/pubsub/google.ts
```diff
@@ -0,0 +1,7 @@
+import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
```
tjmw, line 1: Very minor, but is it worth being explicit that these are types?

```suggestion
import type {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
```

tjmw (approved):
Nice!

cloudformation.yaml
MaelGNM: Will address in a separate PR improving the other descriptions

cloudformation.yaml
MaelGNM: Will address in a separate PR, trying to improve the alarm for all 4 endpoints

typescript/src/feast/pubsub/google.ts
MaelGNM: Good point

cloudformation.yaml
MaelGNM: I think it's worth talking about alarms for this stack in general (especially since they're easy to miss)

PR #1480, opened 2024-05-01
Cloudformation description improvements

MaelGNM:
## What does this change?
This updates descriptions for the pubsub and link endpoint responses as well as for the pubsub 5XX alarms.

I preferred to update the alarms' description rather than changing the threshold as the threshold has previously purposefully been changed to be less sensitive.
We should review all alarms at a later stage.


johnduffell (approved):
nice work thanks for leaving things better than before!

PR #1487, opened 2024-05-03
Handle deleted feast subs

tjmw:
## What does this change?

When a row from the subscriptions Dynamo table is removed, we put an item onto the SQS queue read by the soft-opt-in-consent-setter. When the subscription is a Feast IAP, we should provide a different SOI product name, as the consents are different.

## How to test

We've extended the existing Jest tests and we'll test in CODE.

## Accessibility

### DIFF
--- a/typescript/src/utils/aws.ts
+++ b/typescript/src/utils/aws.ts
@@ -9,7 +9,6 @@ import {PromiseResult} from "aws-sdk/lib/request";
 import SSM = require("aws-sdk/clients/ssm");
 import STS from "aws-sdk/clients/sts";
 import {getMembershipAccountId} from "./guIdentityApi";
-import { Platform } from "../models/platform";
 import { SoftOptInEventProductName } from "./softOptIns";
 
 const credentialProvider = new CredentialProviderChain([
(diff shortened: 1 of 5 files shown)


typescript/src/utils/aws.ts
```diff
@@ -9,6 +9,8 @@ import {PromiseResult} from "aws-sdk/lib/request";
 import SSM = require("aws-sdk/clients/ssm");
 import STS from "aws-sdk/clients/sts";
 import {getMembershipAccountId} from "./guIdentityApi";
+import { Platform } from "../models/platform";
```
MaelGNM, line 12: Is the `Platform` import needed? It doesn't seem to be used

typescript/tests/link/deleteLink.test.ts
```diff
@@ -135,6 +134,45 @@ describe("handler", () => {
         expect(result).toEqual({ recordCount: 1, rowCount: 1 })
     })
 
+    it('puts Feast deletions on the SOI SQS queue the product name FeastInAppPurchase', async () => {
```
MaelGNM, line 137:
```suggestion
    it('puts Feast deletions on the SOI SQS queue with the product name FeastInAppPurchase', async () => {
```

typescript/src/utils/aws.ts
tjmw: Oh good spot! We went through some different refactorings and must have missed this one.

PR #1488, opened 2024-05-03
Use secret in feast google pubsub lambda

MaelGNM:
## What does this change?
This PR:

- adds checking the secret provided in the request to the feast google pubsub lambda
- renames the feast apple pubsub test file
- adds tests for the feast google pubsub test file

## How to test

New tests pass, also tested in CODE.

## Accessibility


typescript/tests/feast/pubsub/google.test.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+beforeEach(() => {
+    process.env['QueueUrl'] = "";
+    process.env['Secret'] = "test_secret";
+});
+
+describe("The Feast Google pubsub", () => {
+    it("Should return HTTP 200", async () => {
+        const input = buildApiGatewayEvent();
```
tjmw, line 55: I think it's _slightly_ unclear, at least at a glance, why this test causes a 200 to be returned and the other a 401. What do you think about making the secret which is used in the request a const which is passed as a parameter to this method? E.g.

```js
const secret = "test_secret";
const input = buildApiGatewayEvent(secret);
```

and in the 401 test:

```
const secret = undefined;
const input = buildApiGatewayEvent(secret);
```

Or similar?

MaelGNM: Good idea!

tjmw (approved):

PR #1511, opened 2024-05-15
Set App tag for alarms

tomrf1:
We're migrating our alarms to google chat.
All of the alarms in this stack should now go to a single sns topic.
The lambda that routes alarms to the relevant chat channel needs an `App` tag on each alarm

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -792,6 +792,9 @@ Resources:
       Statistic: Sum
       Threshold: 1
       TreatMissingData: notBreaching
+      Tags:
+        - Key: App
+          Value: mobile-purchases-google-oauth
 
   GooglePlaySubsStatus5xxErrors:
     Type: AWS::CloudWatch::Alarm
@@ -849,6 +852,9 @@ Resources:
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 1
       Threshold: 20
+      Tags:
+        - Key: App
+          Value: mobile-purchases-google-subscription-status
 
   UpdateGoogleSubscriptionsLambda:
     Type: AWS::Lambda::Function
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -1135,6 +1156,9 @@ Resources:
       OKActions:
         - Ref: AlarmTopic
       TreatMissingData: notBreaching
+      Tags:
+        - Key: App
+          Value: mobile-purchases-apple-pubsub
```
tjmw, line 1161: I _think_ this should have an `App` of `mobile-purchases-apple-update-subscriptions`. That's the lambda which processes the queue and would be the source of any items being added to the dlq, what do you think?

cloudformation.yaml
```diff
@@ -1157,6 +1181,9 @@ Resources:
       OKActions:
         - Ref: AlarmTopic
       TreatMissingData: notBreaching
+      Tags:
+        - Key: App
+          Value: mobile-purchases-google-pubsub
```
tjmw, line 1186: Maybe should be `mobile-purchases-google-update-subscriptions` for the same reasons as the Apple one above?

cloudformation.yaml
```diff
@@ -1206,6 +1233,9 @@ Resources:
       OKActions:
         - Ref: AlarmTopic
       TreatMissingData: notBreaching
+      Tags:
+        - Key: App
+          Value: mobile-purchases-feast-apple-pubsub
```
tjmw, line 1238: Maybe should be `mobile-purchases-feast-apple-update-subscriptions` (I might have the name slightly incorrect here).

tjmw (approved):
Looks really good 👍 

I left a few suggestions in places where I think the app tag should be tweaked slightly, but nothing major (and something we could always fix in a future PR as we get used to this approach).

cloudformation.yaml
tomrf1: You're correct! The relevant bit of the cloudformation is `AppleSubscriptionsEventSource`, which connects `UpdateAppleSubscriptionsLambda` to the queue.

I was thrown by the queue also being passed as an environment variable to `ApplePubSubLambda`

PR #1518, opened 2024-05-22
Fix soft opt-in logging in deleteLink lambda

tjmw:
## What does this change?

Before, the counter for items pushed onto the soft opt-in consent queue was implemented as a global. This meant that it persisted across lambda exectutions resulting in confusing log lines such as:

`Processed 0 records from dynamo stream to disable soft opt-ins for 1 users`.

This moves the counter inside the handler. There's an assumption that if `sendToSqsSoftOptIns` returns a promise which successfully resolves, then the counter can be increased. I think this is fine.

## Accessibility


MaelGNM (approved):
Nice!

PR #1529, opened 2024-06-03
Create `feast-google-update-subscriptions` lambda

michaelbjacobson:
Defines a new lambda for updating Feast subs with the Google Play Store. The lambda's handler function is just a skeleton currently, but it's ready to be built up.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -160,6 +160,7 @@ Resources:
                 - !GetAtt GoogleSubscriptionsQueue.Arn
                 - !GetAtt AppleSubscriptionsQueue.Arn
                 - !GetAtt FeastAppleSubscriptionsQueue.Arn
+                - !GetAtt FeastAndroidSubscriptionsQueue.Arn
                 - !Sub arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:${App}-${Stage}-apple-historical-subscriptions
                 - !Sub arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:${App}-${Stage}-google-historical-subscriptions
         - PolicyName: Kms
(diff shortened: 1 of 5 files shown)


cloudformation.yaml
```diff
@@ -1177,6 +1177,25 @@ Resources:
...
+          App: !Sub ${App}
+          Stack: !Sub ${Stack}
+          Stage: !Sub ${Stage}
+      Description: Consumes Feast subscription data updates from Google Play store from sqs and stores them in dynamo
+      Handler: feast-google-update-subscriptions.handler
+      MemorySize: 512
+      Role: !GetAtt MobilePurchasesLambdasRole.Arn
+      Timeout: 25
+      Runtime: nodejs20.x
+
```
tjmw, line 1198: Should we also add the `EventSourceMapping` here to wire this up with the SQS queue? (Or this could alternatively be done in a follow up PR).

typescript/src/feast/update-subs/google.ts
```diff
@@ -0,0 +1,9 @@
+import { SQSEvent } from "aws-lambda";
+
+export function buildHandler(): (event: SQSEvent) => void {
+    return (event: SQSEvent) => {
+        console.log(`Received SQS event: ${event}`)
```
tjmw, line 5: I _think_ stringifying like this will result in `Received SQS event: [object] Object`. Probably no big deal as this will be short lived but we could either JSON stringify or pass as a second arg to get the full object in the logs:

```
console.log("Received SQS event:", event);
```

What do you reckon?

tjmw (approved):
Looks great! I left a couple of comments to take or leave.

typescript/src/feast/update-subs/google.ts
michaelbjacobson: Good shout, I'll pass it as a second arg.

cloudformation.yaml
michaelbjacobson: I was planning on  doing this in a follow-up, but now I think about it it makes more sense to do here. I'll add it now.

tjmw (approved):
Great!

PR #1530, opened 2024-06-03
initial commit, repurpose existing Google subscription update code fo…

michaelbjacobson:
By making use of existing Google subscription update code from the Live app, this PR implements both the subscription lookup step and the write to Dynamo DB.


MaelGNM (approved):
👍

PR #1536, opened 2024-06-04
Ensure that soft opt-ins are set for Android Feast in-app purchases

michaelbjacobson:
This updates the existing switch statement to ensure that Android Feast IAPs trigger the correct soft opt-ins.

### DIFF
--- a/typescript/src/soft-opt-ins/processSubscription.ts
+++ b/typescript/src/soft-opt-ins/processSubscription.ts
@@ -122,7 +122,7 @@ export async function processAcquisition(subscriptionRecord: ReadSubscription, i
         handleError(`Soft opt-in message send failed for subscriptionId: ${subscriptionId}. ${e}`)
     }
 
-    const isFeast = (): boolean =>
+    const isFeast =
         subscriptionRecord.platform === Platform.IosFeast ||
         subscriptionRecord.platform === Platform.AndroidFeast;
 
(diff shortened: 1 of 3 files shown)


typescript/src/soft-opt-ins/processSubscription.ts
```diff
@@ -122,7 +122,9 @@ export async function processAcquisition(subscriptionRecord: ReadSubscription, i
         handleError(`Soft opt-in message send failed for subscriptionId: ${subscriptionId}. ${e}`)
     }
 
-    const isFeast = subscriptionRecord.platform === Platform.IosFeast;
+    const isFeast =
+        subscriptionRecord.platform === Platform.IosFeast ||
+        subscriptionRecord.platform === Platform.AndroidFeast;
```
MaelGNM, line 127: Technically this change is not needed for the soft opt-in, but is for the email. It's fine to leave it but you'll have a conflict with https://github.com/guardian/mobile-purchases/pull/1535

MaelGNM (approved):
👍

tjmw:
Hi @michaelbjacobson I've merged #1535 this morning which has caused a conflict as @MaelGNM pointed out above. Let me know if I can help resolving! Also in #1535 I added a Jest test for this lambda for an `android-feast` subscription. There's a little bit of the test commented which I think could be useful to test the change you've made in this PR. If you search for `TODO: fix this when correct SOIs are sent` you should find it. Again let me know if I can help!

tjmw (approved):
Nice! 👍

PR #1537, opened 2024-06-05
Fix writing to the soft-opt-in-consent-setter queue in CODE

tjmw:
## What does this change?

Previously running the SOI acquisition lambda in CODE would always fail with an error about writing to the SQS queue for the soft-opt-in-consent-setter lambda. This is because the mobile-purchases codebase expected the soi-consent-setter stage to be `DEV`, but it's actually `CODE`. It was renamed/consolidated to CODE in guardian/support-service-lambdas#1955. Since then, this part of the codebase hasn't worked in mobile-purchases CODE since it was referencing constructs which didn't exist.

## How to test

I've invoked the lambda in CODE and can see that it now successfully writes to the CODE queue for the soft-opt-in-consent-setter.

## Accessibility

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -63,7 +63,7 @@ Mappings:
     CODE:
       Schedule: 'rate(365 days)'
       AlarmActionsEnabled: FALSE
-      SoftOptInConsentSetterStage: DEV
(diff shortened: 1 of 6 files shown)


MaelGNM (approved):
Great to have fixed it!

PR #1545, opened 2024-06-07
Use purchases.subscriptionsv2 for Android Feast subscriptions

tjmw:
## What does this change?

Use the [`purchases.subscriptionsv2`](https://developers.google.com/android-publisher/api-ref/rest/v3/purchases.subscriptionsv2) endpoint to retrieve data about Android Feast subs, instead of [`purchases.subscriptions` v1](https://developers.google.com/android-publisher/api-ref/rest/v3/purchases.subscriptions) (which we use for the live app).

The reason for the migration is that when using the v1 endpoint, we rely on being able to map a product ID to a billing duration. This mapping is defined here:https://github.com/guardian/mobile-purchases/blob/bb1632ea1e53d0fb2ca0e870b67677558a4ea2b9/typescript/src/services/productBillingPeriod.ts#L4. This isn't very robust (in the past, new products have been added in the Play store, but not reflected here, meaning we get rows without a billing period). Furthermore, the was the Feast app has been configured in the Play store, it's not actually possible to map things this way. Instead, if we use the v2 endpoint, we can retrieve the billing period using Play store APIs.

Some back story on the v2 endpont: Tom Wadeson did some work to implement the v2 endpoint. In product there's a test which uses the v2 endpoint in parallel with the v1 endpoint for x% of requests to the subscriptions endpoint. There's a draft PR #1338 to use the new libs everywhere, but not merged yet. Using this for Feast feels like a nice step in the migration path as we're using it for real but in a focused context.

## How to test

I've updated the Jest tests and tested in CODE (and can now see the billing duration being correctly set).

## Accessibility

### DIFF
--- a/typescript/src/feast/update-subs/google.ts
+++ b/typescript/src/feast/update-subs/google.ts
@@ -1,20 +1,48 @@
 import { SQSEvent, SQSRecord } from "aws-lambda";
-import { getGoogleSubResponse } from "../../update-subs/google";
 import { Subscription } from '../../models/subscription';
 import { ProcessingError } from "../../models/processingError";
 import { GracefulProcessingError } from "../../models/GracefulProcessingError";
 import { putSubscription } from "../../update-subs/updatesub";
+import { GoogleSubscription, fetchGoogleSubscriptionV2 } from "../../services/google-play-v2";
+import { GoogleSubscriptionReference } from "../../models/subscriptionReference";
+import { fromGooglePackageName } from "../../services/appToPlatform";
+import { dateToSecondTimestamp, thirtyMonths } from "../../utils/dates";
+
+const googleSubscriptionToSubscription = (
+    purchaseToken: string,
+    packageName: string,
+    googleSubscription: GoogleSubscription
+): Subscription => {
+    return new Subscription(
+        purchaseToken,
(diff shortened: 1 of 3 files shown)


typescript/src/feast/update-subs/google.ts
```diff
@@ -1,20 +1,50 @@
...
-    fetchSubscriptionDetails: (record: SQSRecord) => Promise<Subscription[]>,
+    fetchSubscriptionDetails: (purchaseToken: string, packageName: string) => Promise<GoogleSubscription>,
     putSubscription: (subscription: Subscription) => Promise<Subscription>,
-): (event: SQSEvent) => Promise<string> => (async (event: SQSEvent) => { 
+) => (async (event: SQSEvent) => {
     const promises = event.Records.map(async (sqsRecord: SQSRecord) => {
         try {
-            const subscriptions = await fetchSubscriptionDetails(sqsRecord); // Subscription[]
-            await Promise.all(subscriptions.map(sub => putSubscription(sub)));
+            // TODO: parse this using zod to get validation
```
tjmw, line 39: This TODO is for a future PR.

ecguardian (approved):
👍

MaelGNM (approved):
Looks good!

PR #1554, opened 2024-06-12
Change apple-subscription-status-check-errors alarm

tomrf1:
This endpoint is called by the iOS app to check if a subscription is valid.
If it fails, there is no immediate effect on the user.

This error rate alarm sometimes goes off during quiet periods overnight.
We do get the occasional error, which appears to be on Apple's side. Specifically:
1. Lambda time outs. The timeout is 30 seconds, which is reasonable given this is client facing.
2. `Server error received from Apple, got status 21107, will retry`

This PR:
1. increases the period of the alarm from 10mins to 20mins. This should make it less sensitive to errors during quiet periods.
2. Removes the `OKActions`, which triggers an unhelpful follow up event

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -922,11 +922,9 @@ Resources:
         !FindInMap [StageVariables, !Ref Stage, AlarmActionsEnabled]
       AlarmActions:
         - Ref: AlarmTopic
-      OKActions:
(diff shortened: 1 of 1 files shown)


tjmw (approved):
👍 Looks good!

MaelGNM (approved):
Nice!

PR #1562, opened 2024-06-17
Update the user subscriptions lambda to Node 20

tjmw:
## What does this change?

Update the user subscriptions lambda to Node 20.

## How to test

I've deployed to CODE and seen the lambda successfully invoked.


johnduffell (approved):
looks good, I assume you've tested it or are confident enough of backwards compatibility

PR #1567, opened 2024-06-19
Increase the threshold of Apple Feast sub lookup DLQ alarm

tjmw:
## What does this change?

Increase the threshold of Apple Feast sub lookup DLQ alarm. We seem to have a small trickle of these which fail (which we have an ongoing investigation into) which means this alarm is always firing. I'm concerned that if we pushed out a change which broke _everything_ we wouldn't know about it, as the alarm is already firing. So increase the threshold for now.

## Accessibility


MaelGNM (approved):
👍

PR #1572, opened 2024-06-20
remove unused endpoint in cloudformation.yaml

michaelbjacobson:
Removes unused `/user/subscriptions/me` endpoint.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -266,17 +266,6 @@ Resources:
               responses:
                 "200":
                   "description": "Successfully handled link request from Android Live App"
-          "/user/subscriptions/me":
-            get:
-              x-amazon-apigateway-integration:
-                httpMethod: POST
-                type: aws_proxy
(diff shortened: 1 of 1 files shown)


tjmw (approved):
Looks good 👍 

I _think_ there's also some code that can be removed here as part of removing support for `/user/subscriptions/me`: https://github.com/guardian/mobile-purchases/blob/64b46a3175ebf7248e93ac22119032f0d7c7c84b/typescript/src/user/user.ts#L99 (probably worth double checking, but pretty sure this is the case).

michaelbjacobson:
> Looks good 👍
> 
> I _think_ there's also some code that can be removed here as part of removing support for `/user/subscriptions/me`:
> 
> https://github.com/guardian/mobile-purchases/blob/64b46a3175ebf7248e93ac22119032f0d7c7c84b/typescript/src/user/user.ts#L99
> 
> (probably worth double checking, but pretty sure this is the case).

I think you're right! I had a look at this code to try and find a difference between the two routes, but I totally missed this!

PR #1585, opened 2024-06-27
Prep link endpoint for Feast

tjmw:
## What does this change?

The link endpoint will now handle Feast IAPs (in some cases). When a user takes out an IAP using an offer code, the appAccountToken is not set in the receipt. So we need a different way to link identity IDs with IAPs. We've decided to use the link endpoint used by the live apps to facilitate this. It will only be called for promo code purchases. Other Feast purchases will use the appAccountToken as usual.

**Note about subscription lookups:**

The link endpoint does two things - it creates the user subscription record, and also possibly queues the subscription lookup (pushes onto the SQS queue for the update subs lambda). I don't think this needs to happen for Feast. Furthermore, it's hard to support because the credentials used for lookup in the non-Feast case are different than other apps and in the update subs lambda we don't currently know which app the subscription is for (hence having separate lambdas for Feast and everything else). We _could_ push onto the Feast update subs queue but we know that's going to fail in the user lookup part, due to the aforementioned lack of appAccountToken. So I think the easiest solution for now is just not to enqueue Feast subs for lookup.

## How to test

Tested on CODE. When the platform == `ios-feast` I've seen the user sub record get created, and no lookups queued (as expected). For non `ios-feast` I've seen the sub record created and the lookups queued (existing behaviour for everything other than Feast).

## Accessibility

### DIFF
--- a/typescript/src/link/apple-utils.ts
+++ b/typescript/src/link/apple-utils.ts
@@ -7,7 +7,7 @@ type AppleSubscription = {
 }
 
 export type AppleLinkPayload = {
-    platform: Platform.DailyEdition | Platform.Ios | Platform.IosPuzzles | Platform.IosEdition,
+    platform: Platform.DailyEdition | Platform.Ios | Platform.IosPuzzles | Platform.IosEdition | Platform.IosFeast,
     subscriptions: AppleSubscription[]
 }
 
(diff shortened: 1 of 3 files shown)


ecguardian (approved):
👍

PR #1588, opened 2024-07-01
Set the maximum number of retry events for the SOI acquisitions lambda

tjmw:
## What does this change?

This lambda is triggered by a stream of Dynamo events. We've encountered a scenario where the user deleted their account before this lambda ran, which means we're continually failing to get the user's email address from identity.

If the lambda errors during invocation, we get retries for free: https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html#services-dynamodb-errors

The default retry behaviour is to retry indefinitely: https://docs.aws.amazon.com/lambda/latest/api/API_CreateEventSourceMapping.html (See MaximumRetryAttempts).

There's no point in retrying forever, so set the retry limit to 10 as a starter.

## How to test

- [x] I'll verify the cfn by deploying to CODE.

## Accessibility

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -1477,6 +1477,7 @@ Resources:
             Stream:
               Fn::ImportValue: !Sub ${App}-${Stage}-user-subscriptions-stream-arn
             StartingPosition: LATEST
+            MaximumRetryAttempts: 10
       Tags:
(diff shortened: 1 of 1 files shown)


MaelGNM (approved):
Good fix!

PR #1614, opened 2024-07-30
handle missing userid

Joemgru:
## What does this change?

Occasionally we get responses from Apple and Google that don't have identifiers. Currently the lambda fails and retries. As this is an expected response we should treat it as a success and continue. 

One of the things we do is attempt to read a user identifier from the store receipt data (appAccountToken in the case of Apple, obfuscatedExternalAccountId in the case of Google). If the field doesn’t exist the lambda errors. This results in a number of retries and the event eventually ends up on the DLQ. However, we know there are scenarios where this will happen and are working on mitigations, namely to link to a user another way - using the link endpoint. The iOS fix is already in prod, Android is being worked on.

We therefore shouldn’t treat these as errors as they’re expected, and there’s nothing we can do to fix in this lambda (so retrying makes no sense).

https://trello.com/c/wZrwZqM9/405-mobile-purchases-dont-error-in-the-update-subs-feast-lambdas-if-the-receipt-data-is-missing-a-user-identifier

### DIFF
--- a/typescript/src/feast/update-subs/apple.ts
+++ b/typescript/src/feast/update-subs/apple.ts
@@ -65,7 +65,7 @@ const processRecord = async (
                 const identityId = await exchangeExternalIdForIdentityId(s.appAccountToken)
                 const now = new Date().toISOString()
                 const linked = new UserSubscription(identityId, s.subscriptionId, now)
-                storeUserSubscriptionInDynamo(linked)
+                await storeUserSubscriptionInDynamo(linked)
             }
             else {
                 console.log(`Subscription with receipt '${s.receipt}' did not have an 'appAccountToken'`);
--- a/typescript/src/feast/update-subs/apple.ts
+++ b/typescript/src/feast/update-subs/apple.ts
@@ -62,19 +62,23 @@ const processRecord = async (
 
     await Promise.all(subscriptions.map(storeSubscriptionInDynamo))
 
-    const userSubscriptions =
-        await Promise.all(subscriptions.map(async s => {
-            if (!s.appAccountToken) {
-                throw new ProcessingError(`Subscription with receipt '${s.receipt}' did not have an 'appAccountToken'`, false)
-            }
 
-            const identityId = await exchangeExternalIdForIdentityId(s.appAccountToken)
-            const now = new Date().toISOString()
 
-            return new UserSubscription(identityId, s.subscriptionId, now)
-        }))
-
-    return Promise.all(userSubscriptions.map(storeUserSubscriptionInDynamo))
+            await Promise.all(subscriptions.map(async s => {
+                if (!s.appAccountToken) {
+                    console.log(`Subscription with receipt '${s.receipt}' did not have an 'appAccountToken'`);
+                    return {}
+                    }
+                else {
+                    const identityId = await exchangeExternalIdForIdentityId(s.appAccountToken)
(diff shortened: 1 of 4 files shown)


typescript/src/feast/update-subs/apple.ts
```diff
@@ -54,27 +54,24 @@ const processRecord = async (
...
 
-    const userSubscriptions =
         await Promise.all(subscriptions.map(async s => {
-            if (!s.appAccountToken) {
-                throw new ProcessingError(`Subscription with receipt '${s.receipt}' did not have an 'appAccountToken'`, false)
+            if (s.appAccountToken) {
+                const identityId = await exchangeExternalIdForIdentityId(s.appAccountToken)
+                const now = new Date().toISOString()
+                const linked = new UserSubscription(identityId, s.subscriptionId, now)
+                storeUserSubscriptionInDynamo(linked)
```
tjmw, line 68: I think we should `await` this call. Without it the return value of the promise returned by `storeUserSubscriptionInDynamo` is lost and therefore the outer promise (returned by the `Promise.all` on line 66) may resolve before all `UserSubscription`s have been saved to Dynamo.

typescript/tests/feast/update-subs/apple.test.ts
```diff
@@ -96,14 +95,18 @@ describe("The Feast (Apple) subscription updater", () => {
...
-    })
+
+        await expect(handler(event)).resolves.toBe("OK");
+
+
+        const storedSubscriptionIds =
+            mockStoreSubscriptionInDynamo.mock.calls.map(call => call[0].subscriptionId)
+
+        await expect(mockStoreUserSubscriptionInDynamo.mock.calls.length).toEqual(0)
+        await expect(mockStoreSubscriptionInDynamo.mock.calls.length).toEqual(1)
```
tjmw, line 106: I don't think we need the `await` on these lines (although the mocks return promises, I don't think that's relevant when we're making assertions on them like this).

typescript/tests/feast/update-subs/apple.test.ts
```diff
@@ -96,14 +95,18 @@ describe("The Feast (Apple) subscription updater", () => {
...
+        expect(storedSubscriptionIds).toEqual(["sub-5"])
```
tjmw, line 107: Nice! 👍

typescript/tests/feast/update-subs/google.test.ts
```diff
@@ -71,4 +71,62 @@ describe("The Feast Android subscription updater", () => {
...
+        const expiryTime = plusDays(new Date(), 30);
+        const googleSubscription = {
+            startTime,
+            expiryTime,
+            userCancellationTime: null,
+            autoRenewing: true,
+            productId: subscriptionId,
+            billingPeriodDuration: "P1M",
+            freeTrial: false,
+            testPurchase: false,
```
tjmw, line 94: Just to make the test intent really clear, I wonder if it'd be worth adding:

`obfuscatedExternalAccountId: undefined,`

or similar to highlight the fact that this isn't set. What do you reckon?

typescript/src/feast/update-subs/apple.ts
Joemgru: good spot!

MaelGNM (approved):
Looks good to me!

PR #1654, opened 2024-08-28
Send Feast subscriptions to the "historical" queues

MaelGNM:
## What does this change?

This sends Feast subscriptions to the respective "historical" queues to be exported to GCP.
This changed is necessary as Data Design use data from the "_historical" BigQuery tables.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #1656, opened 2024-08-29
In the Feast google update subs lambda, get the V1 subscription for historical queue

MaelGNM:
## What does this change?
In https://github.com/guardian/mobile-purchases/pull/1654 we started sending Feast subscription to the "historical" queue for them to be recorded in GCP. However for the Google version we use a different version of the API from the live app subscriptions, meaning the data is in a different format and not parsed correctly. Resulting in missing data in the lake.

In this PR we also fetch the V1 subscription specifically to be sent to the "historical" queue.

Longer term we should consider migrating all the lambdas to use the V2 google APIs.

Data for Feast subscriptions is visible in BigQuery.

### DIFF
--- a/typescript/src/feast/update-subs/google.ts
+++ b/typescript/src/feast/update-subs/google.ts
@@ -10,7 +10,8 @@ import { dateToSecondTimestamp, optionalMsToDate, thirtyMonths } from "../../uti
 import { getIdentityIdFromBraze } from "../../services/braze";
 import { storeUserSubscriptionInDynamo, queueHistoricalSubscription } from "./common";
 import { UserSubscription } from "../../models/userSubscription";
-import { fetchGoogleSubscription, GOOGLE_PAYMENT_STATE, GoogleResponseBody } from "../../services/google-play";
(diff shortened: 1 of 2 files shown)


typescript/src/feast/update-subs/google.ts
```diff
@@ -33,8 +34,48 @@ const googleSubscriptionToSubscription = (
     )
 };
 
+const googleResponseBodyToSubscription = (
```
tjmw, line 37: Maybe not for this PR, but would it be worth a follow up to refactor away the duplication between this file and `src/update-subs/google.ts`? It looks like this response parsing code exists in both.

tjmw (approved):
Great 👍

typescript/src/feast/update-subs/google.ts
MaelGNM: Yes, I had considered it but got a bit lost in the code so this way was easier. I'll have a go at refactoring.

tjmw: 👍 Yeah totally understood! As I said I think fine to get this out and attempt the refactor in a follow up PR.

tjmw (approved):

PR #1677, opened 2024-10-14
Set up feast-acquisition-events-router-lambda to pass relevant events to the Apple and Google Processing Lambdas

michaelwmcnamara:
## What does this change?

This work is to configure a lambda that will listen to the event stream for the UserSubscriptions DynamoDB table and whenever a new Feast Event is added to this table, check whether it is android or iOS and pass the UserSubscription record to the corresponding SQS queue.

See the Routing Lambda Function in the following diagram for details:

image

## How to test

Unit Tests to be added
To test manually: 
1 - Deploy to CODE
2 - Add a record to the user-subscriptions Dynamo table with the Platform set to Feast-iOS or Feast-Android
3 - Ensure that the Router lambda is triggered
4 - Ensure that the Subscription Event is sent to the correct Queue
5 - Ensure that the correct processing Lambda is called

This lambda will trigger when a record is added to the User Subscription Dynamo table and pass the corresponding subscription on to the correct Processing Lambda.

### DIFF
--- a/feast-acquisition-events.cloudformation.yaml
+++ b/feast-acquisition-events.cloudformation.yaml
@@ -21,6 +21,7 @@ Parameters:
   AlarmTopic:
     Description: The ARN of the SNS topic to send all the cloudwatch alarms to
     Type: String
+  EventBusName: String
 
 Mappings:
   StageVariables:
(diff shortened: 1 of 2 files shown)


shtukas:
Time to close.

PR #1697, opened 2024-11-12
rename ReadSubscription into SubscriptionEmpty

shtukas:
In this PR we repair an unfortunate misnaming that was introduced 5 years ago, and that created some misunderstanding in the entire code base. 

Let's look at the code at it currently is

```
let itemToQuery = new ReadSubscription();
itemToQuery.setSubscriptionId(subscriptionId);

let subscriptionRecord: ReadSubscription;
subscriptionRecord = await dynamoMapper.get(itemToQuery);
```

In the above sample `itemToQuery` is a `ReadSubscription`. It is not obvious what a `ReadSubscription` is. Is it the type of a subscription returned from the Dynamo Table ?, or the name of a service ? What we know for sure is that it's passed to `dynamoMapper.get`, so it's not a subscription that was just retrieved from the database and not the name of a service.

The story behind `ReadSubscription` is that it's simply a convenience extension of `Subscription` which was introduced to facilitate the instantiation of a new empty Subscription. For this all the fields are set to "" or `undefined` and one method is added to set the important attribute, the `subscriptionId`.

Another problem with the above code is then typing the return value of `dynamoMapper.get` as a `ReadSubscription`, instead of a `Subscription` 

In fact none of the two above imperfections are really a problem, because admittedly a `ReadSubscription` can be used everywhere a `Subscription` can. The problem is that `ReadSubscription` found its way in an unfortunate place, the signature of function with Subscription parameters, for instance 

```
export async function processAcquisition(subscriptionRecord: ReadSubscription, identityId: string): Promise<boolean>
```

To repair this injustice, we rename `ReadSubscription` into `SubscriptionEmpty`, which does a better job at describing that it's a Subscription with zero default values. We also update the signature of `processAcquisition` to be what it should have been 

```
export async function processAcquisition(subscriptionRecord: Subscription, identityId: string): Promise<boolean>
```

And our code sample from the beginning become the much more logical 

```
let itemToQuery = new SubscriptionEmpty();
itemToQuery.setSubscriptionId(subscriptionId);

let subscriptionRecord: Subscription;
subscriptionRecord = await dynamoMapper.get(itemToQuery);
```

Which clearly indicates that we instantiate a `SubscriptionEmpty` of which we are immediately going to update the subscription it, and then that instance of `SubscriptionEmpty` will be used as into of the get function to retrieve a `Subscription`.

### DIFF
--- a/exports-cloudformation.yaml
+++ b/exports-cloudformation.yaml
@@ -116,7 +116,7 @@ Resources:
           Stack: !Sub ${Stack}
           Stage: !Sub ${Stage}
           ExportBucket: !Ref SubscriptionExportBucket
-          ClassName: ReadSubscription
+          ClassName: Subscription
       Description: Export subscription table to the datalake
       MemorySize: 10240
       Timeout: 900
--- a/typescript/src/models/subscription.ts
+++ b/typescript/src/models/subscription.ts
@@ -83,13 +83,13 @@ export class Subscription {
     }
 }
 
-export class ReadSubscription extends Subscription {
+export class SubscriptionEmpty extends Subscription {
 
     constructor() {
         super("", "", "", undefined, false, "", undefined, undefined, undefined)
     }
 
-    setSubscriptionId(subscriptionId: string): ReadSubscription {
+    setSubscriptionId(subscriptionId: string): SubscriptionEmpty {
         this.subscriptionId = subscriptionId;
         return this;
     }
--- a/typescript/src/models/subscription.ts
+++ b/typescript/src/models/subscription.ts
@@ -83,6 +83,13 @@ export class Subscription {
     }
 }
 
+// Note:
+//     SubscriptionEmpty is a convenience class to help with the instanciation of an empty Subscription object.
+//     It's often used to create a argument of dynamoMapper.get (we instanciate it and then
+//     set the subscriptionId, before passing the resulting object to dynamoMapper.get).
+//     It is not meant to stand where a Subscription is the right type, notably as the return type of dynamoMapper.get.
+//     Function dynamoMapper.get will return a Subscription object, not a SubscriptionEmpty object.
+
 export class SubscriptionEmpty extends Subscription {
 
     constructor() {
--- a/typescript/src/export/exportEvents.ts
+++ b/typescript/src/export/exportEvents.ts
@@ -3,7 +3,7 @@ import {dynamoMapper, s3} from "../utils/aws";
 import zlib from 'zlib'
 import {Stage} from "../utils/appIdentity";
 import {DynamoStream} from "./dynamoStream";
-import {ReadSubscriptionEvent, SubscriptionEvent} from "../models/subscriptionEvent";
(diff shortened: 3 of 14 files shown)


exports-cloudformation.yaml
```diff
@@ -116,7 +116,7 @@ Resources:
           Stack: !Sub ${Stack}
           Stage: !Sub ${Stage}
           ExportBucket: !Ref SubscriptionExportBucket
-          ClassName: ReadSubscription
+          ClassName: Subscription
```
shtukas, line 119: See change in typescript/src/export/exportSubscriptions.ts

typescript/src/models/subscription.ts
```diff
@@ -83,20 +83,20 @@ export class Subscription {
...
 }
 
-// TODO: The name "ReadSubscription" is a little bit unfortunate.
-// It's basically an empty subscription that is passed to dynamoMapper.get function
-// Should probably rename it to EmptySubscription.
-// I have noticed that it's miunderstood as a subscription that is being read from the database.
-// For instance in the following code snippet: https://github.com/guardian/mobile-purchases/blob/ccc257c28a7d7a75b9dadfb47f214b074fd8ba50/typescript/src/soft-opt-ins/processSubscription.ts#L101
-// Where in fact that function should take a plain Subscription.
+// Note:
+//     SubscriptionEmpty is a convenience class to help with the instanciation of an empty Subscription object.
```
tjmw, line 87:
```suggestion
//     SubscriptionEmpty is a convenience class to help with the instantiation of an empty Subscription object.
```

typescript/src/models/subscription.ts
```diff
@@ -83,20 +83,20 @@ export class Subscription {
...
+//     It's often used to create a argument of dynamoMapper.get (we instanciate it and then
```
tjmw, line 88:
```suggestion
//     It's often used to create a argument of dynamoMapper.get (we instantiate it and then
```

typescript/src/models/subscription.ts
```diff
@@ -83,20 +83,20 @@ export class Subscription {
...
+//     set the subscriptionId, before passing the resulting object to dynamoMapper.get).
+//     It is not meant to stand where a Subscription is the right type, notably as the return type of dynamoMapper.get.
```
tjmw, line 90: > notably as the return type of dynamoMapper.get

Out of interest, what happens if `dynamoMapper.get` fails to find the subscription? That doesn't result in an empty subscription being returned? I _think_ in this case an error is thrown but just wanted to check.

tjmw (approved):
Looks good! 👍 

(I left a couple of tiny spelling/typo suggestions and a question).

typescript/src/models/subscription.ts
shtukas: That's a good question and no it doesn't return in an instance of `SubscriptionEmpty`, being returned, instead it raises an error and that's the reason `await dynamoMapper.get` is usually surrounded by a `try / catch`

PR #1698, opened 2024-11-12
rename ReadUserSubscription into UserSubscriptionEmpty

shtukas:
This PR is the sister change of the refactoring we applied here: https://github.com/guardian/mobile-purchases/pull/1697

In this case the renaming does an equally good job at highlighting where there was a type confusion, and in particular the  signature of functions that should have taken a `UserSubscription` instead of the old `ReadUserSubscription` are updated. We also update instances of `ReadUserSubscription[]` into `UserSubscription[]`

### DIFF
--- a/exports-cloudformation.yaml
+++ b/exports-cloudformation.yaml
@@ -146,7 +146,7 @@ Resources:
           Stack: !Sub ${Stack}
           Stage: !Sub ${Stage}
           ExportBucket: !Ref UserSubscriptionExportBucket
-          ClassName: ReadUserSubscription
+          ClassName: UserSubscription
       Description: Export the user subscription table to the datalake
       MemorySize: 512
       Timeout: 900
--- a/typescript/src/link/deleteLink.ts
+++ b/typescript/src/link/deleteLink.ts
@@ -12,6 +12,14 @@ async function handleSoftOptInsError(message: string) {
 }
 
 async function getUserLinks(subscriptionId: string) {
+
+    // ( comment group #488db8c1 )
+    // TODO:
+    // In PR: https://github.com/guardian/mobile-purchases/pull/1698
+    // we performed a renaming of ReadSubscription to UserSubscriptionEmpty
+    // With that said it should now be possible to use UserSubscription instead of
+    // UserSubscriptionEmpty as first argument of the dynamoMapper.query(
+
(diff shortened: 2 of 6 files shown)


exports-cloudformation.yaml
```diff
@@ -146,7 +146,7 @@ Resources:
           Stack: !Sub ${Stack}
           Stage: !Sub ${Stage}
           ExportBucket: !Ref UserSubscriptionExportBucket
-          ClassName: ReadUserSubscription
+          ClassName: UserSubscription
```
shtukas, line 149: See change in typescript/src/export/exportSubscriptions.ts

typescript/src/link/deleteLink.ts
```diff
@@ -12,11 +12,11 @@ async function handleSoftOptInsError(message: string) {
 }
 
 async function getUserLinks(subscriptionId: string) {
-    const userLinks = await dynamoMapper.query(ReadUserSubscription, {subscriptionId}, {indexName: "subscriptionId-userId"});
+    const userLinks = await dynamoMapper.query(UserSubscriptionEmpty, {subscriptionId}, {indexName: "subscriptionId-userId"});
```
tjmw, line 15: How is the first argument here used? What does `dynamoMapper.query` do with it?

tjmw (approved):
LGTM! (with a question)

typescript/src/link/deleteLink.ts
shtukas: Good question. The `dynamoMapper.query` method uses this class to understand the structure of the data it needs to query.

tjmw: 👍  So would passing `UserSubscription` work equally well here?

shtukas: @tjmw It would. I didn't do it this time because I didn't want to introduce a change I wasn't 100% sure of (this is only meant to be a renaming), but I have on my list to use `UserSubscription` in those cases, in a future PR

shtukas: @tjmw 
Just in case I forget 😊
https://github.com/guardian/mobile-purchases/pull/1698/commits/173497bf51060f72ce6d6835d9fb22dd69949e8d

PR #1696, opened 2024-11-11
Feast IAP acquisitions: Part 3

shtukas:
Project: Feast subscription data for android and iOS

Part 1: https://github.com/guardian/mobile-purchases/pull/1667
Part 2: https://github.com/guardian/mobile-purchases/pull/1695

Part 3:
In this part we implement the logic of the processing of inserted subscriptions, if they are feast subscriptions.

PR #1695, opened 2024-11-11
Feast IAP acquisitions: Part 2, Introspection of DynamoDBStreamEvents

shtukas:
This is Part 2 (Introspection of DynamoDBStreamEvent for Feast IAP acquisitions), of the publishing of Feast subscription data for android and iOS to the fact_acquisition_events project that started here https://github.com/guardian/mobile-purchases/pull/1667

Here we simply update the primary router to observe DynamoDBStreamEvents.

PR #1667, opened 2024-09-13
Add Feast subscription data to acquisition-event buffer part 1: Add Lambda and SQS cloudformation for apple and google handlers

michaelwmcnamara:
This work is the initial plumbing for two related pieces of work writing feast subscription data for android and iOS to the fact_acquisition_events table via the events buffer. 
Cards for this work: [android](https://trello.com/c/Og1NFpWC/270-feast-iap-acquisitions-should-appear-in-factacquisitionevent) and [iOS](https://trello.com/c/tTfjEXnv/379-feast-iap-acquisitions-should-appear-in-factacquisitionevent).

This change sets up the cloudformation, riffraff etc for this work but does not introduce any functionality.
This work is important because it provides the foundation for two related but independent pieces of work.
Merging this will allow the following work to proceed independently with no risk of overwriting because of changes to the same files

## What does this change?

Add new cloudformation file to handle these new lambda functions and related resources.
This is needed because the existing cloudformation file is at capacity and adding resources their results in cfn template failing validation due to size constraints.

Overall architecture of finished product:

image

Resources added in this PR:
- Feast-acquisition-events router Lambda function
- Feast-acquisition-events processing Lambda functions for apple and google devices
- SQS Queues to communicate between router Lambda function and processing lambdas, corresponding DLQ's and alerts
- IAM Profiles and Policies to enable desired functionality.

Update riffraff.yaml to include references to new cloudformation and lambda functions.
Update ci and webpack files with references to the new resources.

## How to test

- [x] Deploy to CODE
- [x] Ensure the new cloudformation and lambdas have been created.
- [x] Send a message from the apple SQS queue - confirm the apple lambda is called
- [x] Send a message from the google SQS queue - confirm the google lambda is called

##  Note about Deploying for the first time:
It is necessary to perform the following steps to deploy:
1: Preview the deploy, deselect all steps, then select only the steps that upload the lambda files to S3
2: Manually create the mobile-purchases-feast-write-acquisition-events-bus-{STAGE} stack using the feast-acquisition-events.cloudformation.yaml template - being sure to set the correct values for all parameters (Note most of these can be derived from the mobile-purchases stack).
3: Preview the deploy again, deselect all steps then select all steps associated with "mobile-purchases-feast-acquisition-events"

## How can we measure success?

New cloudformation, queues and lambdas have been created and work correctly. These can be used as a basis for further development. 

## Have we considered potential risks?

Minimal risk as no functionality is introduced with these lambdas beyond logging that they have been called.

PR #1700, opened 2024-11-12
Feast IAP acquisitions: Part 4

shtukas:
Project: Feast subscription data for android and iOS

Part 1: https://github.com/guardian/mobile-purchases/pull/1667
Part 2: https://github.com/guardian/mobile-purchases/pull/1695
Part 3: https://github.com/guardian/mobile-purchases/pull/1696

Part 4:
In this part we publish the newly inserted Feast subscriptions to the relevant SQS queue for iOS or Android.

### DIFF
--- a/typescript/src/feast/acquisition-events/router.ts
+++ b/typescript/src/feast/acquisition-events/router.ts
@@ -1,11 +1,51 @@
 import type { DynamoDBRecord, DynamoDBStreamEvent } from 'aws-lambda';
 import { Subscription, ReadSubscription } from "../../models/subscription";
-import { dynamoMapper } from "../../utils/aws";
+import { dynamoMapper, sendToSqs } from "../../utils/aws";
 import { Platform } from "../../models/platform";
+import { plusDays } from "../../utils/dates";
+import { Region, Stage } from "../../utils/appIdentity";
+
+const isActiveSubscription = (currentTime: Date, subscription: Subscription): boolean => {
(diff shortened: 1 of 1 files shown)


typescript/src/feast/acquisition-events/router.ts
```diff
@@ -1,11 +1,47 @@
...
+    // return value indicates whether the processing was successful or not
+    // We return true in the case of an inactive subscription.
+
+    console.log(`[46218776] Processing acquisition for subscription: ${JSON.stringify(subscription)}`);
+    const subscriptionId = subscription.subscriptionId;
+
+    if (!isActiveSubscription(new Date(), subscription)) {
+        console.log(`Subscription ${subscription.subscriptionId} is not active. Processing stopped.`);
+        return true;
+    }
```
MaelGNM, line 27: If the subscription is no longer active (someone took it and immediately cancelled? are there other possible scenarios?), do we not want to record the acquisition anyway?
Is this what we do for other products?

MaelGNM: Thinking about it again, if someone took a subscription and immediately cancelled it, that should be covered by the grace period.
I don't know if there are other cases where we'd see an expired subscription though?

MaelGNM: It's highly unlikely a genuine new acquisition that would be expired would make its way here. So I think it's fine

PR #1704, opened 2024-11-17
Add documentation for promotional offers server side signatures

shtukas:
The implementation of promotional offers server side signatures, which was done more than a year ago, was never documented, and there now is a risk that the knowledge of its existence gets lost. Here we add the documentation for future reference.


shtukas:
Screenshot 2024-11-17 at 22 45 39

PR #1705, opened 2024-11-18
Correct toPlatform function names

shtukas:
This change gives better names to the two functions in appToPlatform.ts. 

They were named using Scala conventions where typically the functions names would be used together with the namespace they belong to. In Scala these two functions would have been defined in a `Platform` object and use as `Platform.fromGooglePackageName`, which is descriptive, but the name `fromGooglePackageName` alone in the middle of TypeScript code is confusing, and even knowing that it was imported from `appToPlatform` doesn't give much intuitive sense of its signature or semantics.

(This is part of the "making the world a better place, one PR at a time" series).


tjmw (approved):
👍

PR #1706, opened 2024-11-18
remove unused function: updateParallelTestTable

shtukas:
Remove an unused function.


tjmw (commented):
This change looks fine to me but might be worth talking to Maël about the plans for this test. It was turned off during the US election as high traffic meant we were hitting our API quota limit and this was an easy way to reduce the number of requests made.

shtukas:
I am closing it now because we might want to actually keep the functionality. I will refactor as part of the Feast integration work I am doing :)

PR #1711, opened 2024-11-20
The interesting story of the Google purchase tokens

shtukas:
Here we add a text file to clarify a piece of idiosyncrasy in the existing code that became apparent during the making of the Google API lookup of the feast integration ( https://github.com/guardian/mobile-purchases/pull/1707 ).

This comes as a separate documentation because although we will be upgrading the pipeline to capture the data that we need to make an Google API call using information contained in a Feast google subscription, the problem around the purchase token and the subscription id in the case of classic google pubsub is not going away anytime soon and this will help future visitors of the code to realize that it's not a problem with them. 

ps: also added a bit of linting and refactoring.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,6 +3,10 @@
 
 mobile-purchases performs at least two primary functions.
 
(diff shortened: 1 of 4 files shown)


shtukas:
Screenshot 2024-11-20 at 13 58 11

MaelGNM (approved):

PR #1713, opened 2024-11-25
Feast IAP acquisitions: Part 6.1, post feast google subscriptions to the acquisition api

shtukas:
Project: Feast subscription data for android and iOS

Part 1: Add Lambda and SQS cloudformation for apple and google handlers: https://github.com/guardian/mobile-purchases/pull/1667
Part 2: Introspection of DynamoDBStreamEvents: https://github.com/guardian/mobile-purchases/pull/1695
Part 3: Feast subscriptions dispatch logic: https://github.com/guardian/mobile-purchases/pull/1696
Part 4: Feast subscriptions SQS insertion: https://github.com/guardian/mobile-purchases/pull/1700
Part 5: Apple and Google API lookups
- 5.1. Introspection: https://github.com/guardian/mobile-purchases/pull/1701
- 5.2. Apple API documentation, and implementation of Apple API look up ( https://github.com/guardian/mobile-purchases/pull/1703 )
- 5.3 Google API documentation, and implementation of Google API look up ( https://github.com/guardian/mobile-purchases/pull/1707 )

Part 6: Post Apple and Google subscriptions to the acquisition API
- 6.1 Google subscriptions

In this change we are completing mobile-purchases-feast-google-acquisition-events by transforming the augmented Google subscriptions to `AcquisitionApiPayload` and posting them to the acquisition API. For the moment we are limiting ourselves only to mandatory attributes.

PR #1730, opened 2024-12-16
determine the paymentFrequency for google subscriptions

shtukas:
Implement the basePlanId to Payment Frequency mapping in the case of Google subscriptions. This is part of the Feast IAP acquisitions projects and second follow up of 6.1 Google subscriptions ( https://github.com/guardian/mobile-purchases/pull/1713 )


MaelGNM (approved):
👍

PR #1742, opened 2025-01-09
Introduce linting

shtukas:
This PR is the first of a few aimed at improving developer experience and here we

1. Upgrade to node 22.13.0 (latest)

2. Use @guardian/prettier ( https://github.com/guardian/csnx/tree/main/libs/@guardian/prettier )

3. Define `.prettierignore` and `.prettierrc.json`

4. Update the package.json's scripts with 

```
"lint": "prettier --check \"./typescript/src/**/*.ts\""
"lint-fix": "prettier --write \"./typescript/src/**/*.ts\""
```


mchv:
A few questions:
- Should you not try using [common prettier config from department](https://github.com/guardian/csnx/tree/main/libs/@guardian/prettier)?
- Why node `21.1.0` and not either `20.18.1` (Iron LTS) or `22.13.0` (Jod LTS)? 

cc @akash1810

shtukas:
Both are good suggestions. Will make a commit. 

About the node version, `21.1.0` was the one suggested by a previous attempt at installing eslint first and then prettier second. After changing my mind and reversing the order (I am going to do eslint in a next PR) that wasn't really relevant anymore. So yes, I can perfectly use the latest version :)

tomrf1 (approved):

PR #1766, opened 2025-02-10
Fix alarm repetition

tomrf1:
We've seen instances of an alarm triggering repeatedly, when actually there was only a single error.
We think it's because of the alarm going into "insufficient data", then looking back at the same period and triggering again.
One solution is to use `FILL` to fill missing data points.

Ref: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#:~:text=A%20particular%20case,use%20shorter%20periods.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -832,7 +832,7 @@ Resources:
       Metrics:
         - Id: e1
           Label: Percentage of requests which result in a 5XX error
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
good!  I edited the description to add the docs link, I hope that's ok, feel free to further edit.

PR #1767, opened 2025-02-10
Fix filled alarm metrics

tomrf1:


johnduffell (approved):
nice one, always a bit harder than you expect to get to the bottom of alarms! 👏

PR #1775, opened 2025-02-17
fix: use log.severe() intead of log.error()

AndreaDiotallevi:
## What does this change?

Fix bug by replacing `log.error` with `log.severe`.

Documentation: https://fivetran.com/docs/connectors/connector-sdk/technical-reference#loggingsyntaxexamples


paulbrown1982 (approved):
👍

PR #1790, opened 2025-03-21
Do not handle missing identityId in braze as an error

tomrf1:
This is to address the alarm:

`mobile-purchases-PROD-FeastAppleSubscriptionDlqDepthAlarm-DpuQdAMu2yds has triggered!`

We sometimes have events that fail because no identity ID exists for a braze user. There is nothing we can do if this happens, so the lambda should just succeed at this point.

Example error:
`ERROR Unexpected error, will throw to retry: Error: Response from Braze for Braze ID '[UUID]' did not contain an identity_id at /var/task/feast-apple-update-subscriptions.js:2:675699`

### DIFF
--- a/typescript/src/feast/update-subs/apple.ts
+++ b/typescript/src/feast/update-subs/apple.ts
@@ -1,12 +1,11 @@
 import type { SQSEvent, SQSRecord } from 'aws-lambda';
 import { App } from '../../models/app';
 import { GracefulProcessingError } from '../../models/GracefulProcessingError';
-import { ProcessingError } from '../../models/processingError';
 import type { Subscription } from '../../models/subscription';
 import type { AppleSubscriptionReference } from '../../models/subscriptionReference';
 import { UserSubscription } from '../../models/userSubscription';
 import { validateReceipt } from '../../services/appleValidateReceipts';
-import { getIdentityIdFromBraze } from '../../services/braze';
+import { getIdentityIdFromBraze, IdentityIdFromBraze } from '../../services/braze';
 import { toAppleSubscription } from '../../update-subs/apple';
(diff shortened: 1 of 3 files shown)


LAKSHMIRPILLAI (approved):
LGTM 👍🏼

typescript/src/feast/update-subs/apple.ts
```diff
@@ -85,12 +84,14 @@ const processRecord = async (
...
     subscriptions.map(async (s) => {
       if (s.appAccountToken) {
-        const identityId = await exchangeExternalIdForIdentityId(
+        const { identityId } = await exchangeExternalIdForIdentityId(
           s.appAccountToken,
         );
-        const now = new Date().toISOString();
-        const linked = new UserSubscription(identityId, s.subscriptionId, now);
-        await storeUserSubscriptionInDynamo(linked);
+        if (identityId) {
```
tjmw, line 90: Possibly as a follow up - would it be useful to have a metric which represents the scenario where there is no `identityId` in Braze? It might be interesting to see the numbers over time.

tjmw (approved):

typescript/src/feast/update-subs/apple.ts
tomrf1: That's a very good point. I will do this, and perhaps at least temporarily we can have an alarm on it so that we can monitor it

tomrf1: @tjmw I've pushed a change to send a metric -
`feast_apple_update_subs_missing_identity_id`
`feast_google_update_subs_missing_identity_id`

tjmw: That's great 👍

PR #1804, opened 2025-04-08
increase acquisition retry time to 8 hours

johnduffell:
We previously limited the retries to stop records retrying for ever in this PR https://github.com/guardian/mobile-purchases/pull/1588

However we recently had an outage due to credential issues, which was resolved but already some records were lost.

In the logs it appeared that there were only 10 retries over the course of a few minutes.

This PR updates it to retry for 8 hours, which should give us time to resolve a simple issue.

There's a good summary of the options available here (assuming it's an asynchronous invocation which I'm not sure about): https://docs.aws.amazon.com/lambda/latest/dg/invocation-async-error-handling.html

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -1479,7 +1479,7 @@ Resources:
             Stream:
               Fn::ImportValue: !Sub ${App}-${Stage}-user-subscriptions-stream-arn
             StartingPosition: LATEST
-            MaximumRetryAttempts: 10
+            MaximumRecordAgeInSeconds: 28800
       Tags:
         Stage: !Ref Stage
         Stack: !Ref Stack


cloudformation.yaml
```diff
@@ -1479,7 +1479,7 @@ Resources:
             Stream:
               Fn::ImportValue: !Sub ${App}-${Stage}-user-subscriptions-stream-arn
             StartingPosition: LATEST
-            MaximumRetryAttempts: 10
+            MaximumRecordAgeInSeconds: 28800
```
johnduffell, line 1482: https://docs.aws.amazon.com/lambda/latest/api/API_CreateEventSourceMapping.html#API_CreateEventSourceMapping_ResponseSyntax:~:text=two%20and%20retry.-,MaximumRecordAgeInSeconds,-%E2%80%93%20Discard%20records%20older

MaelGNM (approved):

johnduffell:
sounds like it will block processing until the error is cleared or the record expires which is not ideal.  It would be good to move to a dead letter queue at some point.

https://docs.aws.amazon.com/lambda/latest/dg/services-dynamodb-errors.html#:~:text=With%20the%20default%20settings%2C%20this%20means%20that%20a%20bad%20record%20can%20block%20processing%20on%20the%20affected%20shard%20for%20up%20to%20one%20day.

PR #1816, opened 2025-04-25
update documentation and add payload logging

shtukas:


paulbrown1982 (approved):
👍 as long as the log volume doesn't explode and the log retention period is still 14 days

PR #1820, opened 2025-04-29
Adding an extra key to subscriptions (Part 1: refactoring)

shtukas:
Over the next few months we are going to perform some important redesign / upgrades / refactoring to mobile-purchases, part of which is to improve data quality. 

As per discussion with DataDesign, the first step of that journey, a temporary measure, is to add a new key to the subscriptions that are sent to 
- datalake.mobile_subscriptions
- datalake.mobile_subscription_events

In this change we update the code layout to help with that aim. This is just a refactoring.


tomrf1 (approved):
👍 looks like it's moving code from `typescript/src/feast/acquisition-events/apple.ts` into some new files

shtukas:
Yeah. Essentially starting to make use of services.

PR #1821, opened 2025-04-29
Adding an extra key to subscriptions (Part 2: more refactoring)

shtukas:
Previous step: https://github.com/guardian/mobile-purchases/pull/1820

In this step: 
- We rename `AppleExtendedData` to the more accurately descriptive `AppleStoreKitSubscriptionDataDerivationForFeastPipeline`
- We introduce `AppleStoreKitSubscriptionData` to represent the data we get from StoreKit.
- We split functions into their basic components and in particular it is now clearly visible that `AppleStoreKitSubscriptionDataDerivationForFeastPipeline` is derived from a `AppleStoreKitSubscriptionData`.
- The function `transactionIdToAppleStoreKitSubscriptionData` is now the center of action as it just takes a `transactionId` and returns a StoreKit data, which is the signature we need for the next step.

### DIFF
--- a/typescript/src/services/api-storekit.ts
+++ b/typescript/src/services/api-storekit.ts
@@ -33,7 +33,7 @@ export interface AppleStoreKitSubscriptionData {
 }
 
 // AppleStoreKitSubscriptionDataDerivation1 is derived from AppleStoreKitSubscriptionData
-export interface AppleStoreKitSubscriptionDataDerivation1 {
+export interface AppleStoreKitSubscriptionDataDerivationForFeastPipeline {
   transactionId: string;
   country: string; // country as two letter code
   currency: string; // currency as three letter code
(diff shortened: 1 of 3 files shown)


typescript/src/services/api-storekit.ts
```diff
@@ -3,10 +3,37 @@ import { forgeStoreKitBearerToken } from './apple-json-web-tokens';
 import { productIdToPaymentFrequency, storefrontToCountry } from './apple-mappings';
 const jwt = require('jsonwebtoken');
 
-// AppleExtendedData is built from the answer from
+// AppleSubscriptionData is built from the answer from
```
tomrf1, line 6: `AppleStoreKitSubscriptionData`?

typescript/src/services/api-storekit.ts
```diff
@@ -3,10 +3,37 @@ import { forgeStoreKitBearerToken } from './apple-json-web-tokens';
...
+  environment: string,
+  transactionReason: string,
+  storefront: string,
+  storefrontId: string,
+  price: number,
+  currency: string,
+  offerDiscountType: string
+}
+
+// AppleStoreKitSubscriptionDataDerivation1 is derived from AppleStoreKitSubscriptionData
```
tomrf1, line 35: old name here

tomrf1 (approved):
just a couple of name issues in the comments

PR #1822, opened 2025-04-29
Adding an extra key to subscriptions (Part 3: upgrade SubscriptionEvent)

shtukas:
Previous step: https://github.com/guardian/mobile-purchases/pull/1821

In this step we introduce `AppleStoreKitSubscriptionDataDerivationForExtra` the official type of the data we are sending to the Lake as additional key to SubscriptionEvent. Each data version is clearly indicated, in this case `apple-extra-2025-04-29`, this will the object to be forward compatible in storage in case DataDesign asks for non backward compatible changes, but also will allow the difference between the apple and google extra data.

And note the decision to update the signature of `toDynamoEvent` in the apple case. This is to preserve testability.

### DIFF
--- a/typescript/src/services/api-storekit.ts
+++ b/typescript/src/services/api-storekit.ts
@@ -61,13 +61,12 @@ interface AppleLatestReceiptInfoItem {
 type AppleLatestReceiptInfo = AppleLatestReceiptInfoItem[];
 
 export const transactionIdToAppleStoreKitSubscriptionData = async (appBundleId: string, transactionId: string): Promise<AppleStoreKitSubscriptionData | null> => {
-  console.log(`[116fa7d4] transactionId: ${transactionId}`);
+  console.log(`[116fa7d4] appBundleId: ${appBundleId}, transactionId: ${transactionId}`);
 
   const url = `https://api.storekit.itunes.apple.com/inApps/v1/subscriptions/${transactionId}`;
   console.log(`[5330931d] url: ${url}`);
 
   const token = await forgeStoreKitBearerToken(appBundleId);
-
   console.log(`[f1335718] ${token}`);
 
   const params = {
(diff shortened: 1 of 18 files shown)


typescript/src/services/api-storekit.ts
```diff
@@ -46,6 +46,38 @@ export interface AppleStoreKitSubscriptionDataDerivationForFeastPipeline {
   // "ANNUALLY"
 }
 
+// AppleStoreKitSubscriptionDataDerivation1 is derived from AppleStoreKitSubscriptionData
+// Was originally introduced as part of adding an extra key to SubscriptionEvent and AppleSubscription
+// That are sent to the Lake.
+// The only difference with AppleStoreKitSubscriptionData is that for safety and forward compatibility
+// we introduce a guVersion, which is going to be incremented if there is any non backward compatible 
+// change in that structure. 
+export interface AppleStoreKitSubscriptionDataDerivationForExtra {
```
tomrf1, line 55: could this be -
`export type AppleStoreKitSubscriptionDataDerivationForExtra = AppleStoreKitSubscriptionData & {  guType: "apple-extra-2025-04-29" }`

typescript/src/services/api-storekit.ts
```diff
@@ -238,3 +270,14 @@ export const appleSubscriptionToAppleStoreKitSubscriptionDataDerivationForFeastP
     paymentFrequency,
   };
 };
+
+export const transactionIdToAppleStoreKitSubscriptionDataDerivation2 = async (transactionId: string): Promise<AppleStoreKitSubscriptionDataDerivationForExtra> => {
```
tomrf1, line 274: old name?

shtukas: Done

tomrf1 (approved):

shtukas:
This is the PR that caused the incident. I am closing it at the state it had when I thought it was working, for archeologists

PR #1830, opened 2025-05-02
upgrade Subscription with a new key: extra

shtukas:
Here we perform for `Subscription` the same operation we did for `SubscriptionEvent` ( https://github.com/guardian/mobile-purchases/pull/1830#issue-3035300883 )

1. Adding an `extra` key 
2. Modify the synchronicity of the corresponding function to enable awaiting for the API answer

### DIFF
--- a/typescript/src/update-subs/apple.ts
+++ b/typescript/src/update-subs/apple.ts
@@ -8,10 +8,11 @@ import { appleBundleToPlatform } from '../services/appToPlatform';
 import { PRODUCT_BILLING_PERIOD } from '../services/productBillingPeriod';
 import { dateToSecondTimestamp, thirtyMonths } from '../utils/dates';
 import { parseAndStoreSubscriptionUpdate } from './updatesub';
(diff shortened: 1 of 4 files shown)


typescript/src/update-subs/apple.ts
```diff
@@ -55,8 +69,8 @@ function sqsRecordToAppleSubscription(
 
   // sandboxRetry is set to false such that in production we don't store any sandbox receipt that would have snuck all the way here
   // In CODE or locally the default endpoint will be sanbox therefore no retry is necessary
-  return validateReceipt(subRef.receipt, { sandboxRetry: false }).then((subs) =>
-    subs.map(toAppleSubscription),
+  return validateReceipt(subRef.receipt, { sandboxRetry: false }).then(async (subs) =>
```
tomrf1, line 72: I don't think the async is necessary here as it's returning the promise

tomrf1 (approved):

PR #1845, opened 2025-05-19
second attempt at prettier

shtukas:
This is a new attempt at activating prettier to this repository, after the first attempt was partially broken (not working for Tom) and also this time was add `prettier:check` to the build itself.

Note that I activated the check but didn't run `prettier:fix` after my "build definition update" commit. This was done on purpose to see if the build would then break and it did

.
Screenshot 2025-05-19 at 12 40 41
.

I then ran `yarn prettier:fix` and we now have a correct build.

The change itself is an update of `package.json` (updating prettier and eslint dependencies) and `build-tc.sh` (adding `prettier:fix`). The other files are just refactoring by `prettier:fix`.


package.json
```diff
@@ -8,8 +8,8 @@
     "build:dev": "webpack",
     "test": "jest",
     "test-lambda": "yarn build:dev && node ./tsc-target/test-launcher.js",
-    "lint": "eslint ./typescript --ext .ts",
```
tomrf1, line 11: I think we still want lint?

shtukas: Ah yeah. I am going to get it back another time, because it flares up everything in that repo (casting to `any` that kind of stuff) and we would need to do some serious refactoring of the code to pass that test.

shtukas: Also, it's currently not part of the build, so it's not exactly a regression :)

tomrf1 (approved):

PR #1894, opened 2025-07-02
Update the retention period to 14 days for all the dead letter queues

LAKSHMIRPILLAI:
## What does this change?

Extend the retention period on mobile purchase SQS

[Trello card](https://trello.com/c/Nyxb3bZ1/1174-extend-the-retention-period-on-mobile-purchase-sqs)

We have encountered several instances where the Mobile Purchase Dead Letter Queue (SQS) alarm is triggered, but upon investigation, the letters in the queue have already disappeared. Additionally, there is no evidence of a purging or triggering of the retry mechanism. This issue occurs because the retention period for letters in the queue is set to 4 days by default.

So we need to update the retention policy for all Mobile Purchase Dead Letter Queues to 14 days to ensure sufficient time for investigation and resolution of issues before letters are removed.

## How to test

Tested in CODE

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -694,6 +694,7 @@ Resources:
     Type: AWS::SQS::Queue
     Properties:
       QueueName: !Sub ${App}-${Stage}-google-subscriptions-to-fetch-dlq
+      MessageRetentionPeriod: 1209600 # 14 days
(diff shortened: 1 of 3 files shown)


sookburt (approved):
Good plan

tomrf1 (approved):
👍

PR #1931, opened 2025-07-28
[extra field project] query the google API for subscriptions

shtukas:
Previously in "Generate Dynamo table `extra` field for subscriptions"...
- Part 1: https://github.com/guardian/mobile-purchases/pull/1921 

----

Here we query the google API and log the information.

### DIFF
--- a/typescript/src/services/google-subscription-extra.ts
+++ b/typescript/src/services/google-subscription-extra.ts
@@ -1,3 +1,6 @@
+import type S3 from 'aws-sdk/clients/s3';
+import aws = require('../utils/aws');
+
 // Author: Pascal
 // This file was introduced in July 2025 to implement the service that queries the Google API
 // to retrieve subscription and subscription product metadata in order to populate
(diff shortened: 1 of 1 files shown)


typescript/src/services/google-subscription-extra.ts
```diff
@@ -1,12 +1,97 @@
...
+    };
+}
+
+export function getAccessToken(stage: string): Promise<AccessToken> {
+    const params: S3.Types.GetObjectRequest = getParams(stage);
+    return aws.s3
+        .getObject(params)
+        .promise()
+        .then((s3OutPut) => {
+            if (s3OutPut.Body) {
```
juabara, line 29: A small suggestion:

```ts
            if (!s3OutPut.Body) {
                throw Error('S3 output body was not defined');                
            }
            return JSON.parse(s3OutPut.Body.toString());
```

juabara (approved):

PR #2130, opened 2026-04-13
remove unused test files

shtukas:
Previously: https://github.com/guardian/mobile-purchases/pull/2129

Seventh episode of a series aimed at removing all linting errors from the entire codebase. Here we remove unused/incomplete test code. This has moved the number of errors from 67 to 64

PR #2131, opened 2026-04-14
linting src/utils/*

shtukas:
Previously: https://github.com/guardian/mobile-purchases/pull/2130

Eigth episode of a series aimed at removing all TypeScript linting errors from the entire codebase. Here we remove linting errors in all the files under src/utils/. This has moved the number of errors from 64 to 54


andrade0 (approved):
LGTM

PR #2145, opened 2026-04-22
move build steps to ci.yaml

shtukas:
Here we get rid of build-tc.sh which was a legacy from the old times of running the builds in TeamCity


.github/workflows/ci.yaml
```diff
@@ -3,112 +3,124 @@ name: CI
...
+          cache: sbt
 
-            - name: Setup Scala
-              uses: guardian/setup-scala@v1
+      - name: Setup Scala
+        uses: guardian/setup-scala@v1
 
-            - name: Test and Build TypeScript
-              run: ./build-tc.sh
+      - name: Test and Build TypeScript
```
shtukas, line 34: Main change happens here, the rest is formatting

graham228221 (approved):

PR #2154, opened 2026-04-27
introduce mpapi-next-gen

shtukas:
introduce mpapi-next-gen

This is a new project inside the existing one; the first step towards a complete upgrade of the existing codebase. The reason for introducing a new package.json is to have a clean set of dependencies. In this change we just introduce a new lambda that is going to eventually replace `mobile-purchases-googleoauth-PROD`. 

This change has no side effects as existing code / functionalities are not changed.


shtukas:
Successfully deployed to CODE 🎉

shtukas:
Screenshot 2026-04-29 at 16 02 34

.github/workflows/ci.yaml
```diff
@@ -46,6 +46,20 @@ jobs:
           do
               zip "${BUNDLE%.*}.zip" "$BUNDLE"
           done
+
+      - name: Build the new lambda collection
+        run: |
+          cd mpapi-next-gen
+          yarn clean
```
johnduffell, line 53: just wondering if we should be looking to move to pnpm for new things?

mpapi-next-gen/src/handlers/mobile-purchases-google-oauth2.ts
```diff
@@ -0,0 +1,7 @@
+//import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3';
```
johnduffell, line 1: needs deleting?

mpapi-next-gen/.gitignore
```diff
@@ -0,0 +1,10 @@
+node_modules/
+dist/
+.env
+*.log
+.DS_Store
+coverage/
+.vscode/
+.idea/
+*.tsbuildinfo
+.DS_Store
```
johnduffell, line 10: DS store is in there twice.  most of these are in the top level gitignore, consider to add the extras (coverage, .env and *.tsbuildinfo if they are needed) there?

mpapi-next-gen/package.json
```diff
@@ -0,0 +1,22 @@
+{
+	"name": "mpapi-next-gen",
```
johnduffell, line 2: would it make sense to call it v2 just in case we end up with a next-next-gen?

mpapi-next-gen/webpack.config.js
```diff
@@ -0,0 +1,41 @@
+const path = require('path');
+
+module.exports = (env = {}) => ({
+	// Good for debugging Lambda errors
+	devtool: 'inline-cheap-source-map',
```
johnduffell, line 5: would be interested to know how it's good - we do something a bit convoluted in support-service-lambdas by producing in the build and attaching  at runtime separate source maps, in order to get proper line numbers in stack traces and logging.  If this is a better way we can switch over accordingly (the support service lambdas way increases cold start time noticable)

cloudformation.yaml
```diff
@@ -394,6 +394,34 @@ Resources:
         Stack: !Ref Stack
         App: !Ref App
 
+  GoogleOAuthLambda2:
```
johnduffell, line 397: out of interest why CFN rather than straight to CDK?  Is that one for future/separate work?

.github/workflows/ci.yaml
shtukas: Good question. This is a huge work in progress and only a first step: I was only interested in solving the deployment process. So I used a template and will address the other things in their separate PRs

mpapi-next-gen/package.json
shtukas: The name is temporary :)

cloudformation.yaml
shtukas: Will do so in due course

andrade0 (approved):
LGTM All good, but a migration to corepack and pnpm would be nice in the future

PR #2157, opened 2026-04-29
upgrade GoogleOAuthLambda runtime to java11

shtukas:
Here we upgrade the java runtime of GoogleOAuthLambda from 8 to 11. This is to allow this change https://github.com/guardian/mobile-purchases/pull/2154 (interestingly the start of the replacement of GoogleOAuthLambda) to deploy correctly


andrade0 (approved):
LGTM

PR #2160, opened 2026-04-29
upgrade apple-fetch-offer-details to node22

shtukas:
Node20 is end of life, here we are upgrading to Node22.


andrade0 (approved):
LGTM

shtukas:
Tested in CODE, same output as in PROD

PR #2161, opened 2026-04-29
upgrade apple-link-user-subscription to node22

shtukas:
Node20 is end of life. Upgrade apple-link-user-subscription to node22


andrade0 (approved):
LGTM

PR #2163, opened 2026-04-29
update node for apple-subscription-status

shtukas:
Node 20 is end of life, this upgrades the runtime for apple-subscription-status to 22


andrade0 (approved):
LGTM

PR #2166, opened 2026-04-29
node22 for delete-user-subscription

shtukas:
Node 20 is end of life, this upgrades delete-user-subscription to Node22.


andrade0 (approved):
LGTM

PR #2196, opened 2026-05-09
Add Bolivia - Bolivian boliviano to country to currency mapping

shtukas:
Will prevent "Error: [acba643d] (countryToCurrencyMap) country BO is not supported", causing an alarm


andrade0 (approved):
LGTM

PR #2208, opened 2026-06-01
refactor acquisitions.ts

shtukas:
While in the process of investigating a problem with acquisitions.ts, we do a bit of refactoring to simplify the existing logic which is needlessly complex. 

We also add markers to the console.log messages for better tracking.


andrade0 (approved):
LGTM

PR #2234, opened 2026-06-18
mpapi-next-gen

shtukas:
Here we set up linting and formatting for `mpapi-next-gen`.

### DIFF
--- a/mpapi-next-gen/.eslintrc.js
+++ b/mpapi-next-gen/.eslintrc.js
@@ -1,0 +1,18 @@
+// .eslintrc.js
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+  },
+  plugins: ['@typescript-eslint'],
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
+  rules: {
+    // Add any custom rules here
+    // You can enable stricter rules over time
+  },
+};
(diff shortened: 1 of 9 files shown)


mpapi-next-gen/.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+// .eslintrc.js
```
tjmw, line 1: Any reason not to use `@guardian/eslint-config`?

mpapi-next-gen/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
tjmw, line 1: Same question about `@guardian/prettier` 😄

mpapi-next-gen/.eslintrc.js
shtukas: Good question. I also have compatibility problems with it, so I will add it in a separate step

mpapi-next-gen/.prettierrc
shtukas: Same answer ☺️

mpapi-next-gen/tsconfig.json
```diff
@@ -1,19 +1,21 @@
...
-		"forceConsistentCasingInFileNames": true,
-		"moduleResolution": "node10",
-		"ignoreDeprecations": "6.0",
-		"types": ["node"],
-		"resolveJsonModule": true
-	},
-	"include": ["src/**/*"],
-	"exclude": ["node_modules", "dist"]
+  "compilerOptions": {
+    "target": "ES2020",
```
tjmw: Could this be something more modern, aligning with the Node runtime we're using?

tjmw (approved):

mpapi-next-gen/tsconfig.json
shtukas: Updated! Thanks!

PR #2235, opened 2026-06-18
use guardian config for eslint and prettier

shtukas:
This is a follow up of https://github.com/guardian/mobile-purchases/pull/2234 , where we start using the guardian specifications for eslint and prettier.


shtukas:
ps: this came in two stages because when I did the same operation with classic/legacy mobile-purchases a few months ago, it was a extremely painful to retroactively get it to work. This time around it went better because the new software is small and self consistent

johnduffell (approved):
nice one, assuming it's mostly just reformatting 👍

rupertbates (approved):
👍

PR #2239, opened 2026-06-18
[mpapi-next-gen]

shtukas:
Here we make another small step in the mpapi-next-gen project by checking that we can write access_token2.json (prototype of access_token.json).


mpapi-next-gen/src/handlers/mobile-purchases-google-oauth2.ts
```diff
@@ -1,6 +1,54 @@
...
-	console.log('[0a433488] new lambda');
+	console.log('[0a433488] starting mobile-purchases-google-oauth2');
+
+	try {
+		// The bucket name is hardcoded in the existing scala code,
+		// so we are doing the same here for simplicity.
+		const bucketName = 'gu-mobile-access-tokens';
+
+		const data = {
+			token: 'fake-token',
```
rupertbates, line 28: This looks like test data?

rupertbates (approved):
👍

mpapi-next-gen/src/handlers/mobile-purchases-google-oauth2.ts
shtukas: Yes it was just that PR, The correct token comes with this one: https://github.com/guardian/mobile-purchases/pull/2241 (just to clarify the `access_token2.json` file is not used, it's just being generated)

PR #2242, opened 2026-06-18
[mpapi-next-gen] switch access token files

shtukas:
At the moment the legacy `access_token.json` file is being generated by the scala code. We recently did some work to have a new TypeScript lambda generating the same data ( https://github.com/guardian/mobile-purchases/pull/2241 ) in the `access_token2.json` file.

Having now confirmed that the new lambda is working fine (the new file is correctly generated every 15 mins), it is now time to let the TS code take over. Here we point the Scala code to `access_token2.json`, and the TS code to `access_token.json`.


shtukas:
Screenshot 2026-06-19 at 11 16 48

tomrf1 (approved):

PR #2243, opened 2026-06-18
update lambda role to adapt to new code conventions

shtukas:
Three updates

1. This is a follow of giving to using the new lambda to retrieve google credentials ( https://github.com/guardian/mobile-purchases/pull/2241 ). Since we are now using a slightly different convention, we need to update the cloudformation

2. We also set the node runtime environment to 22.

3. The code was failing on CODE, while testing, due to problems with the packaging. So we update webpack.


shtukas:
Screenshot 2026-06-18 at 18 25 45

shtukas:
Screenshot 2026-06-18 at 18 26 58

shtukas:
Screenshot 2026-06-18 at 18 28 15

tomrf1 (approved):

PR #2253, opened 2026-06-23
prevent import error in mobile-purchases-googleoauth2

shtukas:
The lambda runtime environment was failing with the following error
```
[Error]: Cannot find module './947.js
```

Here we fix it by moving from webpack to esbuild and setting up the esbuild config correctly. Also note the update in the build script. 

we now have 

```
2026-06-23T12:16:00.107Z	INFO [0a433488] starting mobile-purchases-google-oauth2
2026-06-23T12:16:00.227Z	INFO File uploaded successfully: CODE/google-play-developer-api/access_token.json
2026-06-23T12:16:00.227Z	INFO ETag: "ab33aaf5b50de6cb16b149f8a01503ea"
```


shtukas:
Successfully deployed and tested on CODE

.github/workflows/ci.yaml
```diff
@@ -40,15 +40,31 @@ jobs:
...
-          yarn clean
-          yarn install
-          yarn validate
-          yarn build
-          cd dist
-          for BUNDLE in *.js
-          do
-              zip "${BUNDLE%.*}.zip" "$BUNDLE"
-          done
+            yarn clean
```
johnduffell: possible odd indentation here?  Consider extracting some or all of this section to a ci script

shtukas: Thanks John. Yeah, putting it into a script is my very next step

.github/workflows/ci.yaml
```diff
@@ -40,15 +40,31 @@ jobs:
...
+            yarn install
+            yarn validate
+            yarn build
+            
+            # Package each handler with node_modules
+            cd dist
+            for HANDLER in mobile-purchases-google-oauth2 export-subscription-table-v2 export-historical-data; do
+              # Create temp directory for each handler
```
johnduffell: probably don't need these comments as they just repeat the command's meaning (although a "why" might help on line 48

.github/workflows/ci.yaml
```diff
@@ -45,9 +45,14 @@ jobs:
...
-          for BUNDLE in *.js
-          do
-              zip "${BUNDLE%.*}.zip" "$BUNDLE"
+          for HANDLER in mobile-purchases-google-oauth2 export-subscription-table-v2 export-historical-data; do
+            mkdir -p temp_${HANDLER}
+            cp ${HANDLER}.js temp_${HANDLER}/
+            cp -r ../node_modules temp_${HANDLER}/
+            cd temp_${HANDLER}
+            zip -r "../${HANDLER}.zip" .
+            cd ..
```
johnduffell, line 54: optional: it's more of an fyi but did you know can do pushd and popd to make it slightly more of a "temporary cd"

.github/workflows/ci.yaml
shtukas: Gone!

.github/workflows/ci.yaml
shtukas: Nice! Will try that later

mpapi-next-gen/package.json
```diff
@@ -5,7 +5,9 @@
 	"license": "MIT",
 	"packageManager": "yarn@1.22.22",
 	"scripts": {
-		"build": "webpack --env production",
+		"build": "node esbuild.config.js",
+		"package": "yarn build && cp -r node_modules dist/ && cd dist && zip -r ../lambda-package.zip .",
```
johnduffell, line 9: should this be doing the same as the ci.yaml?  Ideally ci.yaml could call this, then we can easily use it elewhere (e.g. lambda code updater script)

shtukas: Will refactor later

johnduffell (approved):
I can't claim to understand the individual parts, but it looks clear enough overall.

Consider to do a "self review" of PRs like this to flag up areas you are more or less certain about, and provide information that doesn't need to be in the code itself.  The advantage of that is that later when someone wonders why a decision was made, they can go back in history and review the thought process at the time (without it cluttering the code)

Even something like "I've no idea what this file is for but I stole it from Repo X as otherwise it woudln't build" can be as useful for future as "this value is really important otherwise it fails in PROD".

PR #2254, opened 2026-06-23
fix invalid value error in export-google-historical-data

shtukas:
mobile-purchases-export-google-historical-data-PROD is having the following runtime error:

```
Invalid value \"undefined\" for header \"x-amz-decoded-content-length\"
``` 

Here we collect the stream into a buffer before uploading. 

Running on CODE gives the right output:

```
2026-06-23T13:36:16.257Z 6baa35ea-1e18-4341-8005-f9352aab88ae INFO [b3e0e9c1] filename: code-data/date=2026-06-22/2026-06-22-w8c.json.gz
```

But this might not necessarily signal the end of the problem. I need to try in PROD.

### DIFF
--- a/mpapi-next-gen/src/handlers/exportHistoricalData.ts
+++ b/mpapi-next-gen/src/handlers/exportHistoricalData.ts
@@ -153,10 +153,16 @@ export async function handler(params: {
 
 	zippedStream.end();
 
+	const chunks: Buffer[] = [];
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
looks good, as long as the data is a reasonable size (I haven't looked through the rest of the code to see what happened so far)

PR #2255, opened 2026-06-23
flush stuck messages from historical queues

shtukas:
At the moment

Functions
- mobile-purchases-export-apple-historical-data-PROD
- mobile-purchases-export-google-historical-data-PROD

Are reading from SQS queues
- mobile-purchases-PROD-apple-historical-subscriptions
- mobile-purchases-PROD-google-historical-subscriptions

but only reading messages with `snapshotDate` from the date before today. This was based on the assumption that the code would run successfully at least once a day, but that has not been the case and we have 80,000+ messages stuck in the apple queue. This change should take care of that.

Note: no business impact as Data Design no longer use those files

### DIFF
--- a/mpapi-next-gen/docs/mobile-purchases-export-apple-historical-data.md
+++ b/mpapi-next-gen/docs/mobile-purchases-export-apple-historical-data.md
@@ -2,7 +2,8 @@
 export-historical-data writes the datalake files, of the form
 
 ```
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
seems ok although it may be confusing if all the unprocessed data from the previous time ends up filed in the partition for yesterday.. If there aren't too many dates, the other option is to iterate through running the lambda for each, or just flush the SQS queue if we don't need any of the data.

